PR #3, opened 2023-06-30
RD-9092 Implement Http Package in Truffle

Alexzerntev:
- Implemented Http Package
- Added binary const type
- Converted `Exceptions` to `RawTruffleExceptions` , 
- Fixed location settings


raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/http_package/HttpUrlDecodeNode.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
+
+@NodeInfo(shortName = "Http.UrlDecode")
+@NodeChild(value = "str")
+public abstract class HttpUrlDecodeNode extends ExpressionNode {
+    @Specialization
+    public String encode(String str) {
+        return URLDecoder.decode(str, StandardCharsets.UTF_8);
```
miguelbranco80, line 28: No need for Truffle boundaries? Do we have a better understanding of when those are needed?
If this would be part of a larger AST, would it then be necessary?

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/http_package/HttpUrlEncodeNode.java
```diff
@@ -0,0 +1,31 @@
...
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+
+@NodeInfo(shortName = "Http.UrlEncode")
+@NodeChild(value = "str")
+public abstract class HttpUrlEncodeNode extends ExpressionNode {
+
+    @Specialization
+    public String encode(String str) {
+        return URLEncoder.encode(str, StandardCharsets.UTF_8);
```
miguelbranco80, line 29: (Same comment/question regarding Truffle boundaries.)

miguelbranco80: Any need to handle URLEncoder exceptions? Can this fail?
If so @bgaidioz , should it be triable?

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/http_package/HttpReadNode.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            ObjectList headersResult = new ObjectList(headers);
+            records.writeMember(record, "headers", headersResult);
+
+            return ObjectTryable.BuildSuccess(record);
+        } catch (LocationException | UnsupportedMessageException | UnknownIdentifierException |
+                 UnsupportedTypeException |
+                 IOException e) {
+            return ObjectTryable.BuildFailure(e.getMessage());
```
miguelbranco80, line 96: Suggest you distinguish LocationException - whose message is "safe to show" to the users - from all the others, which are internal exceptions whose text we must not share to users for security reasons. In those cases, send a more generic message and log the original exception.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/iterable/collection/CollectionTupleAvgNode.java
```diff
@@ -58,7 +56,8 @@ protected ObjectTryable doCollection(Object iterable,
             }
             records.writeMember(record, "count", results[1]);
             return ObjectTryable.BuildSuccess(record);
-        } catch (Exception ex) {
+        } catch (RawTruffleRuntimeException | UnsupportedMessageException | UnknownIdentifierException |
```
miguelbranco80, line 59: Same comment on exception as early, to distinguish "RawExceptions" whose text is always safe to share, from all others.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/location_package/LocationBuildNode.java
```diff
@@ -74,39 +79,51 @@ public Object executeGeneric(VirtualFrame frame) {
...
-                Object list = listLibs.get(lists, i);
-                Object key = listLibs.get(list, 0);
-                Object val = listLibs.get(list, 1);
-                vec = vec.$plus$eq(Tuple2.apply((String) key, (String) val));
-            }
-            return new raw.sources.LocationKVSetting(vec.result());
-        } else {
-            throw new RawTruffleRuntimeException("Unsupported type for LocationSettingValue", this);
+        } catch (UnsupportedMessageException | UnknownIdentifierException | InvalidArrayIndexException e) {
+            throw new RawTruffleRuntimeException("Error while building LocationSettingValue", this);
```
miguelbranco80, line 126: This message is not helpful to a user. It also talks about internal implementation. Remember that Raw Exceptions become user-visible errors.
Also, it should start with 'small caps' if it were a user-visible error (which it's not)
When can this happen? Is this an assertion error otherwise? I was assuming the type system protects some of this, but may be mistake. Perhaps @bgaidioz can comment further.

miguelbranco80 (changes requested):
Some changes are needed I believe on the error handling part.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/http_package/HttpUrlEncodeNode.java
bgaidioz: I understand `.encode` doesn't fail to encode. It only throws `NullPointerException` when the charset is null. That's why it wasn't made tryable.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/http_package/HttpReadNode.java
Alexzerntev: The other are truffle exceptions of dynamic objects. Aren't they safe to show?

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/location_package/LocationBuildNode.java
bgaidioz: I understand the same: the calls to `InteropLibrary` methods can throw (if one passes an inexisting key to 'readMember', if one calls `readArrayElement` with a wrong index). It is a system error, we have a bug. This should be reported as an `"internal error"` to a user.

In CSV the same happens, when we read into a record or write from a record. I patched the code in an ongoing PR to report `RawTruffleRuntimeException("internal error", cause, this)`.

Maybe we need something consistent to throw all over the place? `throw new RawInternalError(cause, node)`?

Alexzerntev: Makes sense to me

Alexzerntev:
@miguelbranco80 @bgaidioz All the fixes for comments of this PR are fixed in https://github.com/raw-labs/snapi/pull/6 . Which is also merged in this one.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/http_package/HttpReadNode.java
bgaidioz: I understand this is fixed in the latest version of the code (only `LocationException` and `IOException` are caught).

bgaidioz (approved):

miguelbranco80 (changes requested):
Some additional ones.

miguelbranco80 (approved):
Thank you!

PR #6, opened 2023-07-03
RD-9167 make sure that no exception class is used in truffle

Alexzerntev:
- Fixed all occurrences of `Exception` class
- Fixed some warnings

### DIFF
--- a/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/reader/parser/RecordParseCsvNode.java
+++ b/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/reader/parser/RecordParseCsvNode.java
@@ -58,7 +58,7 @@ public class RecordParseCsvNode extends ExpressionNode {
             try {
                 records.writeMember(record, fieldName, value);
             } catch (UnsupportedMessageException | UnknownIdentifierException | UnsupportedTypeException ex) {
-                throw new RawTruffleInternalErrorException(ex.getCause(), this);
+                throw new RawTruffleInternalErrorException(ex, this);
             }
         }
         parser.finishLine(this);
--- a/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/reader/parser/RecordParseCsvNode.java
+++ b/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/reader/parser/RecordParseCsvNode.java
@@ -23,7 +23,7 @@ import raw.compiler.rql2.source.Rql2AttrType;
 import raw.runtime.truffle.ExpressionNode;
 import raw.runtime.truffle.RawLanguage;
 import raw.runtime.truffle.ast.ProgramExpressionNode;
-import raw.runtime.truffle.runtime.exceptions.csv.CsvParserRawTruffleException;
+import raw.runtime.truffle.runtime.exceptions.RawTruffleInternalErrorException;
 import raw.runtime.truffle.runtime.record.RecordObject;
 
 @NodeInfo(shortName = "RecordParseCsv")
@@ -58,7 +58,7 @@ public class RecordParseCsvNode extends ExpressionNode {
             try {
                 records.writeMember(record, fieldName, value);
             } catch (UnsupportedMessageException | UnknownIdentifierException | UnsupportedTypeException ex) {
-                throw new CsvParserRawTruffleException(parser, ex, this);
+                throw new RawTruffleInternalErrorException(ex.getCause(), this);
             }
         }
         parser.finishLine(this);
--- a/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/http_package/HttpReadNode.java
+++ b/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/http_package/HttpReadNode.java
@@ -94,7 +94,7 @@ public abstract class HttpReadNode extends ExpressionNode {
         } catch (LocationException | IOException e) {
             return ObjectTryable.BuildFailure(e.getMessage());
         } catch (UnsupportedMessageException | UnknownIdentifierException | UnsupportedTypeException e) {
-            throw new RawTruffleInternalErrorException(e);
+            throw new RawTruffleInternalErrorException(e, this);
         }
     }
 
--- a/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/http_package/HttpReadNode.java
+++ b/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/http_package/HttpReadNode.java
@@ -24,6 +24,7 @@ import raw.runtime.RuntimeContext;
 import raw.runtime.truffle.ExpressionNode;
 import raw.runtime.truffle.RawContext;
 import raw.runtime.truffle.RawLanguage;
+import raw.runtime.truffle.runtime.exceptions.RawTruffleInternalErrorException;
 import raw.runtime.truffle.runtime.list.ListLibrary;
 import raw.runtime.truffle.runtime.list.ObjectList;
 import raw.runtime.truffle.runtime.option.OptionLibrary;
@@ -90,10 +91,10 @@ public abstract class HttpReadNode extends ExpressionNode {
             records.writeMember(record, "headers", headersResult);
 
             return ObjectTryable.BuildSuccess(record);
-        } catch (LocationException | UnsupportedMessageException | UnknownIdentifierException |
-                 UnsupportedTypeException |
-                 IOException e) {
+        } catch (LocationException | IOException e) {
             return ObjectTryable.BuildFailure(e.getMessage());
+        } catch (UnsupportedMessageException | UnknownIdentifierException | UnsupportedTypeException e) {
+            return ObjectTryable.BuildFailure(RawTruffleInternalErrorException.message);
         }
     }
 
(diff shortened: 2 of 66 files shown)


raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/location_package/LocationDescribeNode.java
```diff
@@ -208,7 +213,8 @@ protected Object doDescribe(LocationObject locationObject, int sampleSize,
 
 
             return ObjectTryable.BuildSuccess(record);
-        } catch (Exception ex) {
+        } catch (InferrerException | UnsupportedMessageException | UnknownIdentifierException |
+                 UnsupportedTypeException ex) {
```
bgaidioz, line 217: Same here I think. The `InferrerException` is safe, but the other three are showing an internal error, right?

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/iterable/list/ListBuildNode.java
```diff
@@ -110,7 +111,7 @@ public Object executeGeneric(VirtualFrame frame) {
                 }
                 return new ObjectList(values);
             }
-        } catch (Exception ex) {
+        } catch (UnexpectedResultException | RawTruffleRuntimeException ex) {
```
bgaidioz, line 114: Here too I feel the `UnexpectedResultException` would be showing an internal error (it's `.executeAsString` throwing it). Is it user-friendly the `.getMessage` of that one?

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/runtime/operators/CompareOperator.java
```diff
@@ -168,7 +169,8 @@ static int doRecord(CompareOperator comparator, RecordObject left, RecordObject
                     }
                 }
                 return 0;
-            } catch (Exception e) {
+            } catch (InvalidArrayIndexException | UnsupportedMessageException | UnknownIdentifierException |
```
bgaidioz, line 172: Again that case of internal error.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/runtime/exceptions/RawTruffleInternalErrorException.java
```diff
@@ -0,0 +1,36 @@
...
+ * licenses/APL.txt.
+ */
+
+package raw.runtime.truffle.runtime.exceptions;
+
+import com.oracle.truffle.api.nodes.Node;
+
+
+// This exception is thrown when an internal error occurs in the Raw Truffle runtime.
+public class RawTruffleInternalErrorException extends RawTruffleRuntimeException {
```
bgaidioz, line 19: I suppose an internal error is meant to be reported to the user as a system failure. If it inherits RawTruffleRuntimeException, is there a risk that it's caught and wrapped into a tryable.error or something like that? Or maybe one needs certain guidelines in tryable related code to make sure this one goes to the top.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/runtime/function/Closure.java
```diff
@@ -42,7 +42,7 @@ public Object call(Object... arguments) {
         try {
             return interop.execute(function, args);
         } catch (UnsupportedTypeException | ArityException | UnsupportedMessageException e) {
-            throw new RawTruffleRuntimeException(e.getMessage(), node);
+            throw new RawTruffleInternalErrorException(e.getCause(), node);
```
bgaidioz, line 45: Why `.getCause`? Isn't it skipping one level?

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/http_package/HttpReadNode.java
```diff
@@ -90,10 +91,10 @@ protected ObjectTryable doRead(LocationObject locationObject,
             records.writeMember(record, "headers", headersResult);
 
             return ObjectTryable.BuildSuccess(record);
-        } catch (LocationException | UnsupportedMessageException | UnknownIdentifierException |
-                 UnsupportedTypeException |
-                 IOException e) {
+        } catch (LocationException | IOException e) {
             return ObjectTryable.BuildFailure(e.getMessage());
+        } catch (UnsupportedMessageException | UnknownIdentifierException | UnsupportedTypeException e) {
+            return ObjectTryable.BuildFailure(RawTruffleInternalErrorException.message);
```
bgaidioz, line 97: I feel here it should throw a new RawTruffleInternalException (like in other places) so that the error isn't wrapped in a tryable, but reported as an internal failure to the user.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/runtime/exceptions/RawTruffleInternalErrorException.java
Alexzerntev: We catch RawTruffleRuntimeException on top (outside truffle). Even if it is wrapped into tryable, it will still reach the top. The only case it won't, is lists, collections and records. How do you think we should proceed?

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/runtime/function/Closure.java
Alexzerntev: Right

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/reader/parser/RecordParseCsvNode.java
```diff
@@ -58,7 +58,7 @@ public Object executeGeneric(VirtualFrame frame) {
             try {
                 records.writeMember(record, fieldName, value);
             } catch (UnsupportedMessageException | UnknownIdentifierException | UnsupportedTypeException ex) {
-                throw new CsvParserRawTruffleException(parser, ex, this);
+                throw new RawTruffleInternalErrorException(ex, this);
```
miguelbranco80, line 61: I'm not sure what's RawTruffleInternalErrorException but sounds more like an assertion than otherwise.
Make sure it does *NOT* inherit from RawTruffleRuntimeException because it is NOT a runtime error - in the sense of user-visible error during runtime.
It looks like it aborts the query "badly". When can this happen? A bug on our side? Then you can either let it go through - the top-level will anyway capture unexpected crashes - or you can wrap it as you do into something of ours, if that makes it more 'visible'.
But is it recoverable from? Does wrapping add any value?

Alexzerntev: In TruffleCompiler.scala we only catch RawTruffleRuntimeException, what will happen if we throw something else?

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/runtime/exceptions/RawTruffleInternalErrorException.java
bgaidioz: Indeed that one shouldn't be caught and reported as a basic error in records and list items. It's an internal error.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/reader/parser/RecordParseCsvNode.java
bgaidioz: These `UnsupportedMessageException` and others aren't recoverable. In this context they indicate a bug on our side.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/string_package/StringReadLinesNode.java
```diff
@@ -35,7 +35,7 @@ protected Object doExecute(LocationObject locationObject, String encoding) {
             TruffleInputStream stream = new TruffleInputStream(locationObject, context);
             TruffleCharInputStream charStream = new TruffleCharInputStream(stream, encoding);
             return new ReadLinesCollection(charStream, context);
-        } catch (Exception ex) {
+        } catch (RawTruffleRuntimeException ex) {
             throw new RawTruffleRuntimeException(ex.getMessage(), this);
```
miguelbranco80, line 39: Catching and rethrowing the same thing...

Alexzerntev: fixed

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/http_package/HttpReadNode.java
```diff
@@ -90,10 +91,10 @@ protected ObjectTryable doRead(LocationObject locationObject,
             records.writeMember(record, "headers", headersResult);
 
             return ObjectTryable.BuildSuccess(record);
-        } catch (LocationException | UnsupportedMessageException | UnknownIdentifierException |
-                 UnsupportedTypeException |
-                 IOException e) {
+        } catch (LocationException | IOException e) {
             return ObjectTryable.BuildFailure(e.getMessage());
+        } catch (UnsupportedMessageException | UnknownIdentifierException | UnsupportedTypeException e) {
+            throw new RawTruffleInternalErrorException(e);
```
bgaidioz: This can embed `this` (the `Node`).

bgaidioz (approved):

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/string_package/StringReadLinesNode.java
bgaidioz: I thought it was to add `this` (the `Node`) to the exception.

PR #7, opened 2023-07-03
add-on to RD-9167: Define a specific CSV writer exception

bgaidioz:
This changeset defines a specific `CsvWriterRawTruffleException` for CSV writing, to be in line with other writers. Before merging make sure all these writer exceptions follow the same scheme (see the PR comment).

### DIFF
--- a/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/writer/internal/RecordWriteCsvNode.java
+++ b/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/writer/internal/RecordWriteCsvNode.java
@@ -24,6 +24,7 @@ import com.oracle.truffle.api.nodes.NodeInfo;
 import raw.runtime.truffle.StatementNode;
 import raw.runtime.truffle.ast.ProgramStatementNode;
 import raw.runtime.truffle.runtime.exceptions.RawTruffleRuntimeException;
+import raw.runtime.truffle.runtime.exceptions.csv.CsvWriterRawTruffleException;
 import raw.runtime.truffle.runtime.record.RecordObject;
 
 import java.io.IOException;
@@ -63,7 +64,7 @@ public class RecordWriteCsvNode extends StatementNode {
             }
             doEndRow(generator);
         } catch (UnsupportedMessageException | InvalidArrayIndexException | UnknownIdentifierException e) {
-            throw new RawTruffleRuntimeException(e.getMessage(), this);
+            throw new RawTruffleRuntimeException("internal error", e, this);
         }
     }
 
@@ -71,8 +72,8 @@ public class RecordWriteCsvNode extends StatementNode {
     private void doStartRow(CsvGenerator gen) {
         try {
             gen.writeStartArray();
-        } catch (IOException ex) {
-            throw new RawTruffleRuntimeException(ex, this);
+        } catch (IOException e) {
+            throw new CsvWriterRawTruffleException(e.getMessage(), e, this);
         }
     }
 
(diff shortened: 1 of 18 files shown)


raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/runtime/exceptions/csv/CsvWriterRawTruffleException.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+package raw.runtime.truffle.runtime.exceptions.csv;
+
+import com.oracle.truffle.api.nodes.Node;
+import raw.runtime.truffle.runtime.exceptions.RawTruffleRuntimeException;
+
+public class CsvWriterRawTruffleException extends RawTruffleRuntimeException {
+  public CsvWriterRawTruffleException(Throwable ex, Node location) {
+    super("Could not write CSV output", ex, location);
```
bgaidioz, line 20: I believe there should be a specific string here. But the Json one doesn't have any. What's the regular shape?

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/runtime/exceptions/csv/CsvWriterRawTruffleException.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+package raw.runtime.truffle.runtime.exceptions.csv;
+
+import com.oracle.truffle.api.nodes.Node;
+import raw.runtime.truffle.runtime.exceptions.RawTruffleRuntimeException;
+
+public class CsvWriterRawTruffleException extends RawTruffleRuntimeException {
+  public CsvWriterRawTruffleException(Throwable ex, Node location) {
+    super("failed to write CSV", ex, location);
```
bgaidioz, line 20: `JsonWriterRawTruffleException` doesn't embed a JSON format-specific string message. What is the regular shape of these exceptions?

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/writer/internal/RecordWriteCsvNode.java
```diff
@@ -63,25 +64,25 @@ public void executeVoid(VirtualFrame frame) {
             }
             doEndRow(generator);
         } catch (UnsupportedMessageException | InvalidArrayIndexException | UnknownIdentifierException e) {
-            throw new RawTruffleRuntimeException(e.getMessage(), this);
+            throw new RawTruffleRuntimeException("internal error", e, this);
```
bgaidioz, line 67: That's an internal error. How should it be reported?

Alexzerntev (approved):

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/writer/internal/RecordWriteCsvNode.java
Alexzerntev: These are InteropLibrary exceptions. Should we hide them?

bgaidioz: In that case, it's a bug (in our code) to hit them right? It means we're picking bad field names, reading too far in an array, etc.

Alexzerntev: Yes

PR #8, opened 2023-07-03
RD-9401 Fix to CLI crash when pressing <TAB>

miguelbranco80:


bgaidioz (approved):
I guess the bottom part is fixing the bug. The rest isn't totally obvious to me.

Alexzerntev (approved):

PR #11, opened 2023-07-03
RD-8998: Support for binary items in JSON and CSV

bgaidioz:
Both JSON and CSV are expected to support binary fields/columns, by turning them into their base64-encoded string.

I've edited existing tests to add a binary value to the output data and checked it's rendered as the base64-encoded string.

The CSV test has some more changes because it was using the wrong ScalaTest predicate.

### DIFF
--- a/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/writer/internal/BinaryWriteCsvNode.java
+++ b/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/writer/internal/BinaryWriteCsvNode.java
@@ -17,6 +17,7 @@ import com.oracle.truffle.api.CompilerDirectives;
 import com.oracle.truffle.api.frame.VirtualFrame;
 import com.oracle.truffle.api.nodes.NodeInfo;
 import raw.runtime.truffle.StatementNode;
+import raw.runtime.truffle.runtime.exceptions.csv.CsvWriterRawTruffleException;
 
 import java.io.IOException;
 import java.util.Base64;
@@ -38,7 +39,7 @@ public class BinaryWriteCsvNode extends StatementNode {
       String result = Base64.getEncoder().encodeToString(value);
       gen.writeString(result);
     } catch (IOException e) {
-      throw new CsvWriterRawTruffleException(e.getMessage());
+      throw new CsvWriterRawTruffleException(e.getMessage(), e, this);
     }
   }
 }
--- a/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/writer/internal/BinaryWriteCsvNode.java
+++ b/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/writer/internal/BinaryWriteCsvNode.java
@@ -17,38 +17,28 @@ import com.oracle.truffle.api.CompilerDirectives;
 import com.oracle.truffle.api.frame.VirtualFrame;
 import com.oracle.truffle.api.nodes.NodeInfo;
 import raw.runtime.truffle.StatementNode;
-import raw.runtime.truffle.runtime.exceptions.csv.CsvWriterRawTruffleException;
-import raw.runtime.truffle.runtime.primitives.TimeObject;
 
 import java.io.IOException;
-import java.time.LocalTime;
-import java.time.format.DateTimeFormatter;
+import java.util.Base64;
 
-@NodeInfo(shortName = "TimeWriteCsv")
-public class TimeWriteCsvNode extends StatementNode {
-
-  private final DateTimeFormatter formatter1 = DateTimeFormatter.ofPattern("HH:mm:ss");
-  private final DateTimeFormatter formatter2 = DateTimeFormatter.ofPattern("HH:mm:ss.SSS");
+@NodeInfo(shortName = "BinaryWriteCsv")
+public class BinaryWriteCsvNode extends StatementNode {
 
   @Override
   public void executeVoid(VirtualFrame frame) {
     Object[] args = frame.getArguments();
-    TimeObject value = (TimeObject) args[0];
+    byte[] value = (byte[]) args[0];
     CsvGenerator generator = (CsvGenerator) args[1];
     doWrite(value, generator);
   }
 
   @CompilerDirectives.TruffleBoundary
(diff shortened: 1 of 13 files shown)


raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/writer/internal/TimestampWriteCsvNode.java
```diff
@@ -19,27 +19,34 @@
...
+  private void doWrite(TimestampObject value, CsvGenerator gen) {
+    try {
+      LocalDateTime ts = value.getTimestamp();
+      if (ts.getNano() != 0) {
+        gen.writeString(formatter2.format(ts));
+      } else {
+        gen.writeString(formatter1.format(ts));
+      }
+    } catch (Exception e) {
+      throw new RuntimeException(e.getMessage());
```
miguelbranco80, line 49: Both the catch Exception and throwing RuntimeException are wrong in this one. I believe you wanted to catch IOException + formatter exceptions, and throw RawRuntimeTruffle?

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/writer/internal/BinaryWriteCsvNode.java
```diff
@@ -0,0 +1,43 @@
...
+    doWrite(value, generator);
+  }
+
+  @CompilerDirectives.TruffleBoundary
+  private void doWrite(byte[] value, CsvGenerator gen) {
+    try {
+      String result = Base64.getEncoder().encodeToString(value);
+      gen.writeString(result);
+    } catch (Exception e) {
+      throw new RuntimeException(e.getMessage());
```
miguelbranco80, line 40: Both the catch Exception and throwing RuntimeException are wrong in this one. I believe you wanted to catch IOException + formatter exceptions, and throw RawRuntimeTruffle?

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/writer/internal/TimeWriteCsvNode.java
```diff
@@ -19,27 +19,34 @@
...
+  private void doWrite(TimeObject value, CsvGenerator gen) {
+    try {
+      LocalTime ts = value.getTime();
+      if (ts.getNano() != 0) {
+        gen.writeString(formatter2.format(ts));
+      } else {
+        gen.writeString(formatter1.format(ts));
+      }
+    } catch (Exception e) {
+      throw new RuntimeException(e.getMessage());
```
miguelbranco80, line 49: Both the catch Exception and throwing RuntimeException are wrong in this one. I believe you wanted to catch IOException + formatter exceptions, and throw RawRuntimeTruffle?

miguelbranco80 (commented):
Some observations on exceptions.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/csv/writer/internal/TimeWriteCsvNode.java
bgaidioz: Yes, waiting for CSV related changes in #3.

bgaidioz:
> Some observations on exceptions.

I think the concerns are addressed. `IOException` is caught and rewritten as `CsvWriter..Exception`.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/json/writer/internal/TimeWriteJsonNode.java
```diff
@@ -21,21 +21,32 @@
...
 @NodeInfo(shortName = "TimeWriteJson")
 public class TimeWriteJsonNode extends StatementNode {
 
-    public void executeVoid(VirtualFrame frame) {
-        Object[] args = frame.getArguments();
-        this.doWrite((TimeObject) args[0], (JsonGenerator) args[1]);
-    }
+  // two different formatters, depending whether there are milliseconds or not
+  private final DateTimeFormatter fmtWithMS = DateTimeFormatter.ofPattern("HH:mm:ss.SSS");
+  private final DateTimeFormatter fmtWithoutMS = DateTimeFormatter.ofPattern("HH:mm:ss");
```
miguelbranco80, line 32: Can you just double-check that using these won't launch any additional specific exceptions? I see below you only capture IOException but are calling fmtWithMS.format for instance.

miguelbranco80 (approved):
Looks good but double-check if date/time involve other exceptions being thrown.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/json/writer/internal/TimeWriteJsonNode.java
bgaidioz: Good point.

Actually `.format` is documenting as possibly throwing a `DateTimeException`. From the code, I see it's if its internal `StringBuilder` throws an `IOException` on `.append`. I'd say we consider it an internal error (?). This has nothing to do with a bad format (which is already checked by `.ofPattern`) or a parse error (here we're not parsing, just rendering a proper time/timestamp).

If so, do we need to catch and throw a `RawTruffleInternalErrorException`?  I think not because it isn't a Truffle framework exception, it can go to the runner.

bgaidioz: Note. This is also valid for DateWrite which wasn't patched in the changeset.

miguelbranco80: If Java is not complaining, then you don't need to catch and rethrow. My understanding (@Alexzerntev can correct) is that the `RawTruffleInternalErrorException` is there as an unchecked exception to be thrown when you have to handle checked exceptions. So it's ok to let it pass as-is.

bgaidioz: Done.

PR #17, opened 2023-07-05
RD-9182 migrate from jjwt-jackson to jwt-core

datYori:
The current jjwt-jackson is causing transitive dep vuln. there is no new version since april 2022.
We decide here to use `jwt-core` from [github.com/jwt-scala](https://github.com/jwt-scala) instead.

### DIFF
--- a/raw-sources-http/src/main/scala/raw/sources/bytestream/http/oauth2clients/GoogleApiKeyOAuth2Client.scala
+++ b/raw-sources-http/src/main/scala/raw/sources/bytestream/http/oauth2clients/GoogleApiKeyOAuth2Client.scala
@@ -16,7 +16,7 @@ import com.fasterxml.jackson.core.JsonProcessingException
 import com.fasterxml.jackson.databind.PropertyNamingStrategy
 import com.fasterxml.jackson.databind.annotation.JsonNaming
 import com.typesafe.scalalogging.StrictLogging
-import io.jsonwebtoken.Jwts
+import pdi.jwt.{Jwt, JwtAlgorithm, JwtClaim}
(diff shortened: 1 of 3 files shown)


raw-sources-http/src/main/scala/raw/sources/bytestream/http/oauth2clients/GoogleApiKeyOAuth2Client.scala
```diff
@@ -79,15 +76,16 @@ class GoogleApiKeyOAuth2Client extends OAuth2Client with StrictLogging {
...
       val kf = KeyFactory.getInstance("RSA");
       val privKey = kf.generatePrivate(keySpec);
+      val currentTime = Instant.now.getEpochSecond
+
+      val claim = JwtClaim()
+        .by(clientEmail)
+        .to(tokenUri)
+        .issuedAt(currentTime)
+        .expiresAt(currentTime + 3600) // 1 hour
+        .+("scope", scope)
```
miguelbranco80, line 86: @torcato , you're the one to review here.

torcato: There is a test that exercises this code in  `TestOauth2Clients`

torcato: I run that test and everything is good.
We can thing about running it in the CI. There is a linkedin test failing though, which I think its known

torcato (approved):
LGTM

PR #20, opened 2023-07-05
RD-9149: RDBMS "Query" extension

bgaidioz:
## Details

* Oracle isn't covered because of the missing .jar file in CI. I haven't tried to test it.
* There is no Teradata package, therefore no testing of that backend.
* Added new tests to MySQL, PostgreSQL, SqlServer and Snowflake that exercise all types and fail to run a query.
* Tests that run `Read` or `InferAndRead` are disabled using `assume(language != "rql2-truffle")`
* Found a bug in `LocationDescribe` (it was reporting the nullability of the record _for each field_).

## Note on error handling and messages.

I added a test that runs a faulty query (the query refers to a table that doesn't exist, and this fails when creating the query at runtime). The error that is reported is the `SQLException` message string given by the RDBMS backend, like we do in the scala executor (I tested the new tests asserting on the error message in Scala and they pass too).

Do we want to rewrite them as ours? For example the inexistent table is reported as:
* PostgreSQL: "`ERROR: relation "..." does not exist\nPosition 15`" (some indication of an error location in the query string)
* MySQL: "`Table '...' doesn't exist`"
* Snowflake: "`SQL compilation error:\nObject '...' does not exist or not authorized.`"
* SqlServer: "`Invalid object name '...'.`"

Browsing the web for a moment I see the exception is internally a vendor-specific one, which if we catch it with the related libraries, can help to forge our version of the error message and perhaps give some standard error string for most errors? At the same time, these error messages are RDBMS errors a user could need to search the web for.


miguelbranco80:
i think it's ok to stick with vendor-specific messages; more so because this is Query call that is itself vendor-specific.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/jdbc/ByteReadJdbcQuery.java
```diff
@@ -0,0 +1,34 @@
...
+@NodeInfo(shortName = "Jdbc.ByteRead")
+public class ByteReadJdbcQuery extends ExpressionNode {
+
+  private final String idx;
+
+  public ByteReadJdbcQuery(String idx) {
+    this.idx = idx;
+  }
+
+  public Object executeGeneric(VirtualFrame frame) {
```
miguelbranco80, line 28: If we are doing specialized calls, shouldn't we be doing executeByte ? Applies to others that refer to primitive types as well.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/jdbc/DoubleReadJdbcQuery.java
```diff
@@ -0,0 +1,34 @@
...
+@NodeInfo(shortName = "Jdbc.DoubleRead")
+public class DoubleReadJdbcQuery extends ExpressionNode {
+
+  private final String idx;
+
+  public DoubleReadJdbcQuery(String idx) {
+    this.idx = idx;
+  }
+
+  public Object executeGeneric(VirtualFrame frame) {
```
miguelbranco80, line 28: Shouldn't we do executeDouble? (See comment above.)

Alexzerntev (approved):

miguelbranco80 (approved):

PR #21, opened 2023-07-05
RD-9188 upgrade jackson-dataformat-yaml

datYori:
The current version is causing a transitive vuln issue (check JIRA for more details)

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -56,18 +56,19 @@ object Dependencies {
 
   val guava = "com.google.guava" % "guava" % "31.1-jre"
 
+  val jacksonDepsVersion = "2.14.2"
   val jacksonDeps = Seq(
-    "com.fasterxml.jackson.core" % "jackson-core" % "2.14.2",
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -56,18 +56,19 @@ object Dependencies {
 
   val guava = "com.google.guava" % "guava" % "31.1-jre"
 
+  val jacksonDepsVersion = "2.14.2"
```
miguelbranco80, line 59: We used to place the versions separately but then Intellij stopped detecting updates. That's why we have it explicit per dependency per line instead of placing the number separately. (It's also more confusing because you updated YAML module of Jackson only.)

I think the correct solution is to leave the numbers per line, but actually update them all to 2.15.2, so that we move all of Jackson forward.

That said, this must be validated against scala2 and repose. That's because in the past, the latest Jackson version caused issues downstream in Repose. (@gkiomour might know the details.) Now they are more split, so perhaps easier. But whatever version we define here affects Repose because Jackson is part of the Snapi clients, and Repose itself depends on the clients, which then bring along all of Jackson.

miguelbranco80 (approved):

PR #25, opened 2023-07-07
chore: use env for google api tests

datYori:
Make use of the existing env variables for google api tests context


torcato (approved):
LGTM

miguelbranco80 (approved):

PR #26, opened 2023-07-07
Fix exception handling for the Compiler API.

miguelbranco80:
Ensure Compiler throws CompilerException or children.
Change test infra and CLI to capture those instead.


bgaidioz (approved):
I tested it successfully against a development branch that asserts a `runErrorAs`.

PR #27, opened 2023-07-07
RD-9224: support for top level methods

bgaidioz:
This is a working version, although not sure of all details. I see it works.

I added the following steps:
* register methods using the existing `recurseFunProto`.
* added a `MethodRefNode` node to implement `IdnExp` when it points to a method (instead of using `ReadLocalVariable` and the like, I think those would't work. That node embeds a ref to the top-level function to be called. It's very similar to a `ClosureNode` except that it doesn't embed a materialized frame.
* `InvokeNode`: handle both objects `Closure` and `MethodRef` using `isInstanceOf`. I thought of defining a specific `MethodInvoke` node but when generating an `InvokeNode`, one doesn't know which kind of object will be passed as a parameter (e.g. `List.Transform` can receive a method or a closure).

I enabled a bunch of regression tests (see the code). By doing that I hit a couple of new bugs. See inline comments.

### DIFF
--- a/raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/Rql2TruffleCompiler.scala
+++ b/raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/Rql2TruffleCompiler.scala
@@ -348,6 +348,7 @@ class TruffleEmitterImpl(tree: Tree)(implicit programContext: ProgramContext)
         case _: Rql2ListType => getFrameDescriptorBuilder.addSlot(FrameSlotKind.Object, getIdnName(entity), null)
         case _: FunType => getFrameDescriptorBuilder.addSlot(FrameSlotKind.Object, getIdnName(entity), null)
         case _: Rql2RecordType => getFrameDescriptorBuilder.addSlot(FrameSlotKind.Object, getIdnName(entity), null)
+        case _: Rql2LocationType => getFrameDescriptorBuilder.addSlot(FrameSlotKind.Object, getIdnName(entity), null)
       }
       addSlot(entity, slot)
       WriteLocalVariableNodeGen.create(recurseExp(e), slot, rql2Type)
--- a/raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/Rql2TruffleCompiler.scala
+++ b/raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/Rql2TruffleCompiler.scala
@@ -218,12 +218,16 @@ class Rql2TruffleCompiler(implicit compilerContext: CompilerContext)
           )
         )
       case "binary" =>
-        val writer = TruffleBinaryWriter(tree.analyzer.tipe(me.get).asInstanceOf[Rql2BinaryType])
-        new ProgramStatementNode(
-          RawLanguage.getCurrentContext.getLanguage,
-          frameDescriptor,
-          new BinaryWriterNode(bodyExpNode, writer)
-        )
+        tree.analyzer.tipe(me.get) match {
+          case binary: Rql2BinaryType => // nullable or tryable is OK
+            val writer = TruffleBinaryWriter(binary)
+            new ProgramStatementNode(
+              RawLanguage.getCurrentContext.getLanguage,
+              frameDescriptor,
+              new BinaryWriterNode(bodyExpNode, writer)
+            )
+          case _ => throw new CompilerException("unsupported type")
+        }
       case null | "" => new ProgramExpressionNode(
           RawLanguage.getCurrentContext.getLanguage,
           frameDescriptor,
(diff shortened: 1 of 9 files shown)


raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/function/InvokeNode.java
```diff
@@ -39,13 +40,24 @@ public InvokeNode(ExpressionNode functionNode, ExpressionNode[] argumentNodes) {
     public Object executeGeneric(VirtualFrame frame) {
         CompilerAsserts.compilationConstant(argumentNodes.length);
 
-        Closure closure = (Closure) functionNode.executeGeneric(frame);
-
-        Object[] argumentValues = new Object[argumentNodes.length];
-        for (int i = 0; i < argumentNodes.length; i++) {
-            argumentValues[i] = argumentNodes[i].executeGeneric(frame);
+        Object executable = functionNode.executeGeneric(frame);
```
bgaidioz, line 43: I don't think one can even specialize. Methods and functions are compatible (`FunType`). There could be a list containing both, and each is passed as a parameter to `Invoke`.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/iterable/list/ListTakeNode.java
```diff
@@ -128,7 +128,7 @@ protected ObjectList doObject(Object list, long num, @CachedLibrary("list") List
         if (num >= innerList.length) {
             return (ObjectList) list;
         }
-        Object[] result = new String[(int) num];
+        Object[] result = new Object[(int) num];
```
bgaidioz, line 131: A copy/paste bug I noticed after enabling one of the tests.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/json/reader/TryableTopLevelWrapper.java
```diff
@@ -29,21 +27,21 @@
 public class TryableTopLevelWrapper extends ExpressionNode {
 
     @Child
-    private DirectCallNode childDirectCall;
+    private ExpressionNode reader;
```
bgaidioz, line 30: The fix to RD-9228. The nested reader isn't necessarily a parameterless top-level program.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/function/InvokeNode.java
bgaidioz: Actually, that means absolutely all 'callers' (`Invoke` but also `Transform`, `Unnest`, `Filter`, etc.) should have a similar fix.

No test is exercising that these guys support both closures and methods. I open a JIRA for that (it's not a hard problem but this should be fixed after top-level methods are supported). https://raw-labs.atlassian.net/browse/RD-9231 (has some more comments on the problem.)

raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/Rql2TruffleCompiler.scala
```diff
@@ -218,12 +218,16 @@ class Rql2TruffleCompiler(implicit compilerContext: CompilerContext)
           )
         )
       case "binary" =>
-        val writer = TruffleBinaryWriter(tree.analyzer.tipe(me.get).asInstanceOf[Rql2BinaryType])
-        new ProgramStatementNode(
-          RawLanguage.getCurrentContext.getLanguage,
-          frameDescriptor,
-          new BinaryWriterNode(bodyExpNode, writer)
-        )
+        tree.analyzer.tipe(me.get) match {
```
bgaidioz: Protecting 'binary' but we need to protect all (https://raw-labs.atlassian.net/browse/RD-9229).

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/json/reader/TryableTopLevelWrapper.java
bgaidioz: See the ScalaTest I added for it.

miguelbranco80 (approved):

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/function/MethodRefNode.java
```diff
@@ -0,0 +1,35 @@
...
+    @CompilationFinal
+    private final Function function;
+
+    public MethodRefNode(Function f) {
+        this.function = f;
+    }
+
+    @Override
+    public Object executeGeneric(VirtualFrame virtualFrame) {
+        return new MethodRef(this.function, this);
```
Alexzerntev, line 33: I'm a little confused here, since the MethodRef is called with VirtualFrame in any case. Why don't we make it closure (with VirtualFrame instead of MaterializedFrame - so the Frame will do) ? @miguelbranco80

PR #31, opened 2023-07-11
add raw sources tests to ci

datYori:
we add to the CI the ability to check if there is any changes in folder like `raw-source*` and run the associated tests if needed.
To accomplish this we go through the changes between PR and target branch ( filetring `raw-sources*`), then and we run `sbt test` for each subproject containing a change.
We also decides that if there is changes only inside `raw-sources*` we don't run `truffle-tests`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,12 +12,12 @@ env:
   SBT_OPTS : -Dsbt.log.noformat=true -Xss2m -Xms1g
 
 jobs:
-  raw-sources-changed-files:
+  changed-files:
     runs-on: self-hosted
     outputs:
       all_sbt_projects: ${{ steps.convert.outputs.all_sbt_projects }}
-      any_changed: ${{ steps.rawsources.outputs.any_changed }}
-      only_changed: ${{ steps.rawsources.outputs.only_changed }}
+      sources_any_changed: ${{ steps.rawsources.outputs.any_changed }}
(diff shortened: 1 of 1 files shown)


datYori:
I will remove the fakefiles before merging, they are there to give you a sense on what the CI results will look like when changes are occuring in `raw-source*`

miguelbranco80:
>  We also decides that if there is changes only inside raw-sources* we don't run truffle-tests

Question as I don't recall the root issue: what was the exact issue and what is special about raw sources that requires filtering?
A change in raw-sources should trigger all tests to run since truffle depends on raw sources. A new feature or change of behaviour means all tests (truffle or otherwise) should run.

Normally I'd expect raw sources to just trigger all tests as normally and no special flows needed, since raw-sources is another normal package that is needed by truffle, like all others.

datYori:
You are right on triggering all tests even if a change occurs only on `raw-source*`, I ll correct that.
However for now the thing is we only run tests with the tag `@TruffleTests`, which are exclusively located in `raw-compiler-rql2-truffle` 
Meaning we are leaving behind some tests like 
`raw-sources-http/src/test/scala/raw/sources/bytestream/TestHttpServer.scala`
So the idea is to say if a change occured in one of the `raw-sources*` projects, lets run the associated unit tests. ( concept that should be extended to projects like `raw-inferrer-local`)
Another idea would be to replace our specific `sbt "testOnly * -- -n raw.testing.tags.TruffleTests"` test command to just test everything each time

PR #32, opened 2023-07-11
RD-8272 implement json infer and parse

Alexzerntev:
- Added `Json.InferAndParse` . For this entry a new sources package had to be built called *raw-sources-in-memory* . It implements ByteStreamLocation traits and instead of reading from URL like other Location objects, it reads directly from memory. The schema for it is called `in-memory:`.
- Fixed some warnings
- Removed deprecated sparkUrl field in Locations

### DIFF
--- a/raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/builtin/JsonPackageTest.scala
+++ b/raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/builtin/JsonPackageTest.scala
@@ -80,18 +80,49 @@ trait JsonPackageTest extends CompilerTestContext with FailAfterNServer {
 
   private val recordData = tempFile("""{"a": 1, "b": 10, "c": 100}""")
 
-  test(rql"""Json.Read("$data", type collection(undefined))""")(it => it should evaluateTo("""[
-    |  Error.Build("expected null but got non-null"),
-    |  Error.Build("expected null but got non-null"),
-    |  Error.Build("expected null but got non-null")
-    |] """.stripMargin))
+  test(rql"""Json.Read("$data", type collection(undefined))""")(it => it should evaluateTo(
+    """[
+      |  Error.Build("expected null but got non-null"),
+      |  Error.Build("expected null but got non-null"),
+      |  Error.Build("expected null but got non-null")
+      |] """.stripMargin))
 
   test(rql"""Json.Read("$tryNothing1", type collection(undefined))""")(it =>
     it should evaluateTo(""" [null, null, null, null, null, Error.Build("expected null but got non-null")] """)
   )
 
   test(rql"""Json.Read("$tryNothing2", type collection(record(a: int, b: int, c: undefined)))""")(it =>
-    it should evaluateTo("""
+    it should evaluateTo(
+      """
+        |[
+        |  {a: 3, b: 30, c: null},
+        |  {a: 3, b: 30, c: null},
+        |  {a: 3, b: 30, c: null},
+        |  {a: 3, b: 30, c: null},
+        |  {a: 3, b: 30, c: null},
+        |  {a: 3, b: 30, c: Error.Build("expected null but got non-null")}
+        |]""".stripMargin)
+  )
+
+  test(rql"""Json.Read("$tryNothing3", type collection(record(a: int, b: int, c: list(undefined))))""")(it =>
+    it should evaluateTo(
+      """
+        |[
+        |  {a: 3, b: 30, c: []},
+        |  {a: 3, b: 30, c: []},
+        |  {a: 3, b: 30, c: []},
+        |  {a: 3, b: 30, c: []},
+        |  {a: 3, b: 30, c: []},
+        |  {a: 3, b: 30, c: [Error.Build("expected null but got non-null"), null]}
+        |]""".stripMargin)
+  )
+
+  test(rql"""Json.InferAndRead("$tryNothing1", sampleSize = 4)""")(it =>
+    it should evaluateTo("""[null, null, null, null, null, Error.Build("expected null but got non-null")]""")
+  )
+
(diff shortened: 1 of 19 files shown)


raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/builtin/JsonPackageTest.scala
```diff
@@ -209,6 +211,12 @@ trait JsonPackageTest extends CompilerTestContext with FailAfterNServer {
     |  Json.Parse("[1,2,3]", type t)
     |""".stripMargin)(it => it should typeAs("collection(int)"))
 
+  test("""
+    |let t = type collection(int)
+    |in
+    |  Json.InferAndParse("[1,2,3]")
+    |""".stripMargin)(it => it should typeAs("collection(int)"))
+
```
miguelbranco80, line 219: Suggest adding a few more tests to cover types.
Also, suggest you add a test of a non-JSON document, to see how the error reporting looks like. Use `should runErrorAs(<error message>)`.

raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/builtin/JsonPackageTest.scala
```diff
@@ -276,7 +284,7 @@ trait JsonPackageTest extends CompilerTestContext with FailAfterNServer {
     _ should runErrorAs("path not found")
   )
 
-  val jsonWithNulls = tempFile(
+  val jsonWithNulls: Path = tempFile(
```
miguelbranco80, line 287: Suggest you make these vals private instead of annotating the type.

raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/builtin/JsonPackageTest.scala
```diff
@@ -338,7 +346,7 @@ trait JsonPackageTest extends CompilerTestContext with FailAfterNServer {
         |]""".stripMargin)
   )
 
-  val changeTypes = tempFile(
+  val changeTypes: Path = tempFile(
```
miguelbranco80, line 349: Suggest you make these vals private instead of annotating the type.

raw-sources-dropbox/src/main/scala/raw/sources/filesystem/dropbox/DropboxPath.scala
```diff
@@ -30,8 +30,6 @@ class DropboxPath(
 
   override def rawUri: String = s"dropbox://${cli.name}$path"
 
-  override def sparkUri: String = throw new ByteStreamException("spark url not supported for Dropbox")
```
miguelbranco80, line 33: Thank you for cleaning this up!

raw-sources-in-memory/src/main/scala/raw/sources/bytestream/in_memory/InMemoryByteStreamLocation.scala
```diff
@@ -0,0 +1,64 @@
...
+  SeekableInputStream
+}
+
+import java.io.{ByteArrayInputStream, InputStream}
+import java.nio.file.Path
+
+object InMemoryByteStreamLocation {
+  val schema = "in-memory"
+  val schemaWithColon = s"$schema:"
+  val codeDataKey = "code-data"
```
miguelbranco80, line 29: Suggest calling it "data" instead?

raw-sources-in-memory/src/main/scala/raw/sources/bytestream/in_memory/InMemoryByteStreamLocation.scala
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  override def getLocalPath(): Path = throw new AssertionError("Calling path on in memory location")
+
+  override def cacheStrategy: CacheStrategy = CacheStrategy.NoCache
+
+  override def retryStrategy: RetryStrategy = NoRetry()
+
+  override def rawUri: String = "in-memory:"
```
miguelbranco80, line 61: I'm torn here. I know we discussed it and agreed, but I'm wondering if we shouldn't instead have in-memory be a base64 string encoding of a binary? In that case, the data would be in the URL itself. It's probably silly and slower so will leave it at your discretion.
If we'd like to expose "in-memory" locations to users, how would that look like?

raw-sources-in-memory/src/main/scala/raw/sources/bytestream/in_memory/InMemoryByteStreamLocation.scala
Alexzerntev: I thought it is more self explanatory like this, there are data in your snapi code so `code-data`.

raw-sources-in-memory/src/main/scala/raw/sources/bytestream/in_memory/InMemoryByteStreamLocation.scala
Alexzerntev: _I'm wondering if we shouldn't instead have in-memory be a base64 string encoding of a binary_
Seems the same to me, but if the string is large enough maybe we will have extra encode-decode overhead.

_If we'd like to expose "in-memory" locations to users, how would that look like?_
`JsonInferAndRead("in-memory://" + Base64.Encode("data"))`

`JsonInferAndRead(InMemoryLocation.Build(string data, string encoding))`

Something like that? But seems a little complex.

raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/builtin/JsonPackageTest.scala
Alexzerntev: I added broken json test, is there any other case that it can fail?

miguelbranco80 (approved):

PR #37, opened 2023-07-12
more consistent folder names

datYori:
I found it silly that `raw-sources` folder was not following the naming convention of all other projects.
same for `raw-compiler`

also `rawSourcesMsSQL` is not the real camel case of `raw-sources-mssql`


miguelbranco80:
The folder names follow the package name in Scala which is "raw.sources".
Adding .api to the package name would mean the sub-packages (e.g. raw.sources.http) now would become siblings not children of the base API.

datYori:
would it be ok go the other way round then = using `rawSources` instead of `rawSourcesApi` in `build.sbt` ?

PR #46, opened 2023-07-18
RD-9240 Address temporary credentials in AWS library

torcato:
I enabled back the aws truffle tests. ignored for the moment the ones that use xml as it is not implemented in truffle.

One thing to discuss before merging how to generate in the CI the environment variables for the credentials 
```
"AWS_SESSION_TOKEN"
"AWS_ACCESS_KEY_ID"
"AWS_SECRET_ACCESS_KEY"
```

A script that I used to generate these (Pavlos gracefully gave it to me). Maybe it helps
```
#!/bin/bash
role_name=Developers
account_id=<your account id>
aws sso login
file=$(find ~/.aws/sso/cache/ -type f -printf "%t - %p\n" | sort -nr | tail -1 | awk '{print $7}')
access_token=$(cat  $file| jq -r .accessToken)
rc=$(aws sso get-role-credentials --account-id $account_id --role-name $role_name --region eu-west-1 --access-token $access_token | jq .)
export AWS_ACCESS_KEY_ID=$(echo $rc | jq -r .roleCredentials.accessKeyId)
export AWS_SECRET_ACCESS_KEY=$(echo $rc | jq -r .roleCredentials.secretAccessKey)
export AWS_SESSION_TOKEN=$(echo $rc | jq -r .roleCredentials.sessionToken)
```

### DIFF
--- a/raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/builtin/TruffleAwsPackage.scala
+++ b/raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/builtin/TruffleAwsPackage.scala
@@ -64,7 +64,19 @@ class AwsV4SignedRequestEntry extends AwsV4SignedRequest with TruffleEntryExtens
 
     val region = maybeRegion.getOrElse(new StringNode("us-east-1"))
 
-    AwsV4SignedRequestNodeGen.create(key, secretKey, service, region, sessionToken, path, method, host, body, urlParams, headers)
+    AwsV4SignedRequestNodeGen.create(
+      key,
+      secretKey,
+      service,
+      region,
+      sessionToken,
+      path,
+      method,
+      host,
+      body,
+      urlParams,
+      headers
+    )
   }
 
 }
--- a/raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/builtin/TruffleAwsPackage.scala
+++ b/raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/builtin/TruffleAwsPackage.scala
@@ -14,13 +14,16 @@ package raw.compiler.rql2.truffle.builtin
 
 import raw.compiler.base.source.Type
 import raw.compiler.rql2.builtin.AwsV4SignedRequest
-import raw.compiler.rql2.source.{Rql2ListType, Rql2RecordType}
+import raw.compiler.rql2.source.{Rql2AttrType, Rql2ListType, Rql2RecordType, Rql2StringType}
 import raw.compiler.rql2.truffle.{TruffleArg, TruffleEntryExtension}
 import raw.runtime.truffle.ExpressionNode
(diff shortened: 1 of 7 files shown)


raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/extensions/AwsPackageTest.scala
```diff
@@ -20,12 +20,19 @@ trait AwsPackageTest extends CompilerTestContext {
...
+  val awsSecretAccessKey = sys.env("AWS_SECRET_ACCESS_KEY")
+
+  secret(authorizedUser, "temporary-key", awsAccessKeyId)
+  secret(authorizedUser, "temporary-secret-key", awsSecretAccessKey)
+  secret(authorizedUser, "session-token", awsSessionToken)
+
   val triple = "\"\"\""
 
-  test("""let data = Xml.Read(
+  ignore("""let data = Xml.Read(
```
Alexzerntev, line 35: They will be also ignored for scala. I think Ben was adding a condition to run tests per case.

torcato: Hello I added a method `amlImplemented` that needs to be overridden by the scala or truffle implementation of the test to make the tests run in scala an not in truffle

Alexzerntev (approved):

raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/builtin/TruffleAwsPackage.scala
```diff
@@ -57,11 +57,26 @@ class AwsV4SignedRequestEntry extends AwsV4SignedRequest with TruffleEntryExtens
...
-      .getOrElse(new ListBuildNode(Rql2ListType(Rql2RecordType(Vector.empty)), Array()))
+      .getOrElse(new ListBuildNode(Rql2ListType(Rql2RecordType(Vector())), Array()))
+
+    val sessionToken = maybeSessionToken
+      .getOrElse(new StringNode(""))
 
     val region = maybeRegion.getOrElse(new StringNode("us-east-1"))
 
-    AwsV4SignedRequestNodeGen.create(key, secretKey, service, region, path, method, host, body, urlParams, headers)
+    AwsV4SignedRequestNodeGen.create(
```
datYori, line 67: why forcing people to have SSO creds, sso is usually for human users, non human users are fine using the classic key and secret key

torcato: The old credentials (non SSO ones) still work this just adds one more option.
This was a feature requested by Pavlos.

I suppose because of the SSO they have to use it like this.
But anyway some users might want to use these temporary credentials anyway so its not bad

torcato:
After a quick discussion with Yann I removed the test and the temporary credentials until further discussion .

The feature stays so calls with these type of authentication will work.

PR #47, opened 2023-07-19
RD-9292 + RD-9293 + RD-9245 automatic github release

datYori:
This PR implement the following
- RD-9292 create github release after a new tag is pushed
- RD-9190 notify main repo when a new version is published
- RD-9245 build and attach `raw-cli-assembly` jar to the release as an asset


torcato (approved):
Cannot comment a lot on these things.
Seems reasonable:
* checking out
* building
* releasing

Approved!

PR #51, opened 2023-07-24
RD-9312 transitive vuln netty-handler

datYori:
This dep was causing transitive vuln issue. See JIRA for more details.

This dep is not used anymore anyway

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -492,7 +492,6 @@ lazy val rawCompilerRql2Truffle = (project in file("raw-compiler-rql2-truffle"))
   )
   .settings(
     buildSettings, // TODO (msb): Promote this to strictBuildSettings and add bail-out annotations as needed,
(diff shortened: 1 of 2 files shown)


miguelbranco80:
I'm a bit confused on this one; I assumed we had migrated to v2 of the AWS SDK but apparently we are still on v1?

miguelbranco80:
I thought v2 was used for the AWS package. This is now done (and tested) in Truffle, right, @Alexzerntev ?

datYori:
this one
https://github.com/raw-labs/snapi/blob/main/project/Dependencies.scala#L15
we are using it for the S3 Filesystem package but the one we delete here is not used

miguelbranco80 (approved):
Understood. So the S3 package uses old v1 (as before) and the AWS package no longer depends on the AWS SDK. Looks ok to me.

PR #55, opened 2023-07-26
RD-9079: Records with duplicate fields names

bgaidioz:
## Main change

The `RecordObject` class has changed:
* values are stored in a `DynamicObject` like they used by stored in, but keys are field _indexes_ instead of field names. That permits to use that class as a storage object without messing up with duplicated keys.
* keys are stored in a `String[]` (possibly duplicated),
* the API write keys takes a key name _and_ an index, that permits to enter duplicated keys.
* the interop API that returns keys (non-duplicated) implements the renaming before returning the list of keys,

The export to JSON is using that API to write a JSON file (since duplicated keys are discouraged by the JSON specs) while certain language nodes (like `Concat`) are using the real keys in order to maintain the proper key names.

The `RecordObject` used to extend `DynamicObject` and have the same API to insert key/values. Here it has a custom API which ideally should be used internally, while using the `Interop` API should be left to external systems. New APIs permit to write keys at a certain index, or for convenience without specifying an index, in which case the 'index++' is used. It also has APIs to read by index (which are now used by nodes that may hit duplicate keys, like `Concat`).

The `RecordObject` still exports the _Interop API_ and several components in the code are accessing it through it. That permits to export it to other Truffle languages. Likely it will be seen as an object with fields that have the keys' names. The `Node` related internal code of _our_ language should use the internal API, so that it can deal with effective field names there's no confusion of which keys are used. This has been done in certain nodes that handle keys in a critical way (`RemoveField`, `Concat`, etc.) but can be implemented in a further patch.

As I was changing the `RecordObject`, I thought since now its usage of a `DynamicObject`  is internal, we can move the boilerplate `Shape` computation to that class.

## Other changes
* `RecordWriteJson` doesn't need the `fieldNameMap` (I think), since fields are written in order of the `RecordType` attributes.
* the `Rql2OutputTestContest` now turns records into sequences of key/values, instead of map. Initially it was to support possible duplicated keys, but they are renamed

### DIFF
--- a/raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/regressions/RD5714Test.scala
+++ b/raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/regressions/RD5714Test.scala
@@ -23,12 +23,7 @@ trait RD5714Test extends CompilerTestContext {
     |    colB = [{id: 2, firstName: "john"}],
     |    join = List.Join(colA, colB, i -> i.name == i.firstName)
     |in Json.Print(join)""".stripMargin) { it =>
-    if (language == "rql2-truffle") {
-      // rql2-truffle fixes duplicated fields as part of RD-9079
-      it should evaluateTo(""" "[{\"id\":1,\"name\":\"john\",\"id\":2,\"firstName\":\"john\"}]" """.stripMargin)
-    } else {
-      it should evaluateTo(""" "[{\"id\":1,\"name\":\"john\",\"id_1\":2,\"firstName\":\"john\"}]" """.stripMargin)
-    }
+    it should evaluateTo(""" "[{\"id\":1,\"name\":\"john\",\"id_1\":2,\"firstName\":\"john\"}]" """.stripMargin)
   }
 
   test("""let colA = [{id: 1, name: "john"}],
@@ -45,11 +40,7 @@ trait RD5714Test extends CompilerTestContext {
         } finally {
           source.close()
         }
-      if (language == "rql2-truffle") {
-        assert(s == """[{"id":1,"name":"john","id":2,"firstName":"john"}]""")
-      } else {
-        assert(s == """[{"id":1,"name":"john","id_1":2,"firstName":"john"}]""")
-      }
+      assert(s == """[{"id":1,"name":"john","id_1":2,"firstName":"john"}]""")
     } finally {
       Files.delete(path)
     }
--- a/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/RawLanguage.java
+++ b/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/RawLanguage.java
@@ -15,7 +15,6 @@ package raw.runtime.truffle;
 import com.oracle.truffle.api.TruffleLanguage;
 import com.oracle.truffle.api.instrumentation.ProvidedTags;
 import com.oracle.truffle.api.instrumentation.StandardTags;
-import com.oracle.truffle.api.object.Shape;
 import com.oracle.truffle.api.nodes.Node;
 import raw.runtime.truffle.runtime.record.RecordObject;
 
@@ -26,12 +25,6 @@ public final class RawLanguage extends TruffleLanguage<RawContext> {
     public static final String ID = "rql";
     public static final String VERSION = "0.10";
     public static final String MIME_TYPE = "application/x-rql";
-
-    private final Shape rootRecordShape;
-
-    public RawLanguage() {
-        this.rootRecordShape = Shape.newBuilder().layout(RecordObject.class).build();
-    }
 
     @Override
     protected final RawContext createContext(Env env) {
@@ -50,11 +43,7 @@ public final class RawLanguage extends TruffleLanguage<RawContext> {
     }
 
     public RecordObject createRecord() {
-        return new RecordObject(rootRecordShape);
-    }
-
-    public RecordObject createRecord(Shape shape) {
-        return new RecordObject(shape);
+        return new RecordObject();
     }
 
 }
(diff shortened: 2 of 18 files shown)


raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/regressions/RD5714Test.scala
```diff
@@ -23,7 +23,7 @@ trait RD5714Test extends CompilerTestContext {
     |    colB = [{id: 2, firstName: "john"}],
     |    join = List.Join(colA, colB, i -> i.name == i.firstName)
     |in Json.Print(join)""".stripMargin)(
-    _ should evaluateTo(""" "[{\"id\":1,\"name\":\"john\",\"id\":2,\"firstName\":\"john\"}]" """.stripMargin)
+    _ should evaluateTo(""" "[{\"id\":1,\"name\":\"john\",\"id_1\":2,\"firstName\":\"john\"}]" """.stripMargin)
```
Alexzerntev, line 26: Will it break scala after release?

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/RawLanguage.java
```diff
@@ -27,12 +26,6 @@ public final class RawLanguage extends TruffleLanguage<RawContext> {
     public static final String VERSION = "0.10";
     public static final String MIME_TYPE = "application/x-rql";
 
-    private final Shape rootRecordShape;
-
-    public RawLanguage() {
-        this.rootRecordShape = Shape.newBuilder().layout(RecordObject.class).build();
```
Alexzerntev, line 33: In the documentation it says that all dynamic object should be created with the same shape. So the internal lookup for values is faster. So this code should exist somewhere globally. (As it here)

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/record/RecordProjNode.java
```diff
@@ -14,39 +14,33 @@
...
 
 @NodeInfo(shortName = "Record.Project")
 @NodeChild("receiverNode")
 @NodeChild("indexNode")
 public abstract class RecordProjNode extends ExpressionNode {
 
     @Specialization(limit = "3")
-    protected Object readMember(Object record, String key,
-                                @CachedLibrary("record") InteropLibrary records) {
+    protected Object readMember(Object record, String key) {
```
Alexzerntev, line 30: Could be RecordObject instead of Object to avoid cast

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/record/RecordProjNode.java
```diff
@@ -14,39 +14,33 @@
...
             throw new RawTruffleInternalErrorException(e, this);
         }
     }
 
-    @Specialization(limit = "3")
-    protected Object readMember(Object record, int index,
-                                @CachedLibrary("record") InteropLibrary records,
-                                @CachedLibrary(limit = "1") InteropLibrary libraries) {
+    @Specialization
+    protected Object readMember(Object rec, int index) {
```
Alexzerntev, line 39: same here

raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/regressions/RD5714Test.scala
bgaidioz: Yes. That test probably fails in Scala. I'm not too scared because we should hit a CI/test error when trying to upgrade the dependency. But yes.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/runtime/record/RecordObject.java
```diff
@@ -114,25 +153,62 @@ boolean isMemberRemovable(String member,
...
+            distinctKeys = null;
+        }
+        keys.set(idx, key);
+        valuesLibrary.put(values, idx, value);
+    }
+
+    // adds a value by key only (auto-increment the index)
+    // TODO replace all internal calls to writeMember by calls to addByKey
+    public void addByKey(String key, Object value) {
+        valuesLibrary.put(values, keys.size(), value); // "key" to use in the dynamic object is the current index.
```
torcato, line 200: Remember to add a jira issue for this. Replacing all writeMember by addByKey

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/runtime/record/RecordObject.java
```diff
@@ -114,25 +153,62 @@ boolean isMemberRemovable(String member,
...
+        keys.add(key); // the original key is added (possible duplicate)
+    }
+
     @ExportMessage
-    public void writeMember(String name, Object value,
-                            @CachedLibrary("this") DynamicObjectLibrary objectLibrary) {
-        objectLibrary.put(this, name, value);
+    // TODO replace all internal calls to writeMember by calls to addByKey
+    public void writeMember(String name, Object value) {
+        addByKey(name, value);
```
torcato, line 207: When we replace the internal usage of writeMember by addByKey then this function can be updated with  the typical behavior.

torcato (approved):
We will have a later change replacing the internal use of writeMember with addByKey and making the writeMember have the typical behavior, overwrite values instead of appending. Maybe something to discuss.

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/runtime/record/RecordObject.java
bgaidioz: https://raw-labs.atlassian.net/browse/RD-9392

PR #61, opened 2023-08-07
RD-9229: data type checks in data exports

bgaidioz:
This is the patch to make sure a query isn't run when its result should be exported to a format that doesn't support the data type (e.g. records with nested collections when exporting to CSV, but also functions, packages, when exporting to anything).

Related packages `JsonPackage`, `CsvPackage`, `BinaryPackage` and `StringPackage` are exposing an API that checks the target data type and returns true/false depending how it's supported. The `Rql2TruffleCompiler` calls these APIs before building the full tree. Once the tree is built, it's wrapped into the writing nodes. I removed some checks that were performed there since we're checking upfront.

I added a test suite to exercise the change, and some existing tests were enabled.

### DIFF
--- a/raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/Rql2TruffleCompiler.scala
+++ b/raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/Rql2TruffleCompiler.scala
@@ -24,9 +24,9 @@ import raw.compiler.base.{BaseTree, CompilerContext}
 import raw.compiler.common.source.{Exp, IdnExp, IdnUse, SourceNode, SourceProgram}
 import raw.compiler.rql2.Rql2TypeUtils.removeProp
 import raw.compiler.rql2._
-import raw.compiler.rql2.builtin.EnvironmentPackageBuilder
+import raw.compiler.rql2.builtin.{BinaryPackage, CsvPackage, EnvironmentPackageBuilder, JsonPackage, StringPackage}
 import raw.compiler.rql2.source._
-import raw.compiler.rql2.truffle.builtin.{CsvWriter, JsonRecurse, TruffleBinaryWriter}
+import raw.compiler.rql2.truffle.builtin.{CsvWriter, JsonIO, TruffleBinaryWriter}
 import raw.compiler.truffle.{TruffleCompiler, TruffleEntrypoint}
 import raw.compiler.{base, CompilerException, ErrorMessage, ProgramSettings}
 import raw.runtime.{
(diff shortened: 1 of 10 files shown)


raw-compiler-rql2/src/main/scala/raw/compiler/rql2/Simplifier.scala
```diff
@@ -0,0 +1,67 @@
...
+import org.bitbucket.inkytonik.kiama.rewriting.Cloner.attempt
+import org.bitbucket.inkytonik.kiama.rewriting.Rewriter._
+import org.bitbucket.inkytonik.kiama.rewriting.Strategy
+import org.bitbucket.inkytonik.kiama.util.Entity
+import raw.compiler.base.Phase
+import raw.compiler.common.source.{Exp, IdnDef, IdnExp, IdnUse, SourceProgram}
+import raw.compiler.rql2.builtin.{CollectionPackageBuilder, ListPackageBuilder}
+import raw.compiler.rql2.source._
+
+class Simplifier(protected val parent: Phase[SourceProgram], protected val phaseName: String)(
```
miguelbranco80, line 24: Can you add a descriptive comment of this phase?

raw-compiler-rql2/src/main/scala/raw/compiler/rql2/Simplifier.scala
```diff
@@ -0,0 +1,67 @@
...
+) extends PipelinedPhase {
+
+  override protected def execute(program: SourceProgram): SourceProgram = {
+    simplify(program)
+  }
+
+  private def simplify(program: SourceProgram): SourceProgram = {
+    val tree = new Tree(program)
+    lazy val analyzer = tree.analyzer
+    val s: Strategy = sometd(rule[Exp] {
```
miguelbranco80, line 35: Descriptive names instead of s and s2

raw-compiler-rql2/src/main/scala/raw/compiler/rql2/Simplifier.scala
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    val s3 = rule[Exp] {
+      case let: Let if let.decls.exists(useless) =>
+        val decls = let.decls.filterNot(useless)
+        if (decls.isEmpty) let.e
+        else Let(decls, let.e)
+    }
+    val r = rewrite(everywhere(s3))(noRefs.root)
+    logger.debug("Simplifier:\n" + format(r))
```
miguelbranco80, line 63: Probably want to remove this one.

raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/Rql2TruffleCompiler.scala
```diff
@@ -236,15 +236,26 @@ class Rql2TruffleCompiler(implicit compilerContext: CompilerContext)
...
 
     val tree = new Tree(program)(programContext.asInstanceOf[ProgramContext])
     val emitter = new TruffleEmitterImpl(tree)(programContext.asInstanceOf[ProgramContext])
-
     val Rql2Program(methods, me) = tree.root
+    val dataType = tree.analyzer.tipe(me.get)
+    val outputFormat = programContext.settings.getString(ProgramSettings.output_format)
+    outputFormat match {
+      case "csv" => if (!CsvWriter.supported(dataType)) throw new CompilerException("unsupported type")
+      case "json" => if (!JsonRecurse.supported(dataType)) throw new CompilerException("unsupported type")
```
miguelbranco80, line 252: We should rename "JsonRecurse" to something sensible...

raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/Rql2TruffleCompiler.scala
```diff
@@ -236,15 +236,26 @@ class Rql2TruffleCompiler(implicit compilerContext: CompilerContext)
...
+      case "text" => if (!dataType.isInstanceOf[Rql2StringType]) throw new CompilerException("unsupported type")
```
miguelbranco80, line 253: What about null types?
Shouldn't we have the check in the respective package, as in CSV and JSON?

bgaidioz: Null types are supported by String and Binary writes. I'm changing the code to have these checks defined in packages. We'll see how that feels.

raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/Rql2TruffleCompiler.scala
bgaidioz: I removed the 'supported' method from that helper object.

bgaidioz: and I renamed it `JsonIO`

miguelbranco80 (approved):

PR #97, opened 2023-08-17
RD-9091: Xml reader

bgaidioz:


raw-compiler-rql2/src/main/scala/raw/compiler/rql2/builtin/XmlPackage.scala
Alexzerntev, line 1: Do we need to think about Xml.InferAndParse now? Or better add it afterwards?

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/io/xml/parser/NullableParseXmlNode.java
Alexzerntev, line 1: Should it be deleted? I think it is a copy paste that is not used

raw-compiler-rql2/src/main/scala/raw/compiler/rql2/builtin/XmlPackage.scala
bgaidioz: Yes we probably need to think about it and implement it. I opened https://raw-labs.atlassian.net/browse/RD-9463 to remember it.

Alexzerntev (approved):

Alexzerntev:
Do new tests pass in scala?

bgaidioz:
> Do new tests pass in scala?

Yes I did test manually locally. But why not testing it once again.

PR #102, opened 2023-08-21
RD-9442 + Truffle's CSV query renders `\N` as `\\N` instead of `N` in Scala

Alexzerntev:
- Added escape char handling to CSV parser which defaults to "\"
- Added test for it
- Formatting

### DIFF
--- a/raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/builtin/CsvPackageTest.scala
+++ b/raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/builtin/CsvPackageTest.scala
@@ -17,10 +17,15 @@ import raw.compiler.rql2.tests.{CompilerTestContext, FailAfterNServer}
 
 trait CsvPackageTest extends CompilerTestContext with FailAfterNServer {
 
+  val ttt = "\"\"\""
+
   private val data = tempFile("""a|b|c
     |1|10|100
     |2|20|200
     |3|30|300""".stripMargin)
+
+  private val dataWithEscaped = tempFile("""a|b|c
+    |1|10|\N""".stripMargin)
 
   private val headerLessData = tempFile("""1|10|100
     |2|20|200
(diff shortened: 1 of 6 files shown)


Alexzerntev:
Formatting shadows the changes, please check anything with "escape" and "escExp" (tried to be consistent) for review.

raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/builtin/CsvPackageTest.scala
```diff
@@ -140,10 +145,8 @@ trait CsvPackageTest extends CompilerTestContext with FailAfterNServer {
     |{3, 30, 300}
     |]""".stripMargin))
 
-  test(rql"""Csv.InferAndRead("$data")""".stripMargin)(it => it should evaluateTo("""[
-    |{a: 1, b: 10, c: 100},
-    |{a: 2, b: 20, c: 200},
-    |{a: 3, b: 30, c: 300}
+  test(rql"""Csv.InferAndRead("$dataWithEscaped")""".stripMargin)(it => it should evaluateTo("""[
```
bgaidioz, line 148: Nice, this is indeed the bug we hit.

I believe we should add:

This to make sure \\ is the default:
```scala
rql"""Csv.Read("$dataWithEscaped", type ..., header, etc.)""" // but no escape
```

This to check the parameter is used
```scala
rql"""Csv.Read("$dataWithEscaped", ...., escape "\\")"""
```

This to make sure both implementations are consistently taking that. Which I believe means "no escape"
```scala
rql"""Csv.Read("$dataWithEscaped", ...., escape = null)"""
```

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/io/csv/reader/parser/RawTruffleCsvParser.java
```diff
@@ -44,603 +44,608 @@
...
+            this.dateFormatter = DateTimeFormatCache.get(settings.dateFormat);
+            this.timeFormatter = DateTimeFormatCache.get(settings.timeFormat);
+            this.timestampFormatter = DateTimeFormatCache.get(settings.timestampFormat);
+            CsvFactory csvFactory = new CsvFactory();
+            csvFactory.enable(CsvParser.Feature.TRIM_SPACES);
+            Reader reader = stream.getReader();
+            jacksonParser = csvFactory.createParser(reader);
+            CsvSchema.Builder builder = CsvSchema.builder();
+            builder.setColumnSeparator(settings.delimiter);
+            if (settings.useQuote) {
```
bgaidioz, line 74: I imagine that works, assuming `useQuote` is set to false when passed as `null`?

Alexzerntev:
> Do you know if these new tests are passing in Scala?

Yes

PR #104, opened 2023-08-22
build: remove unused guava dependency

datYori:
As explained here, we are not using `guava` anymore

https://github.com/raw-labs/snapi/blob/9fada2b665b78cccb1a9dc94b5cd60ffb4de7296/raw-utils/src/main/scala/raw/utils/RawConcurrentHashMap.scala#L22-L24


datYori:
we actually still use it in raw-utils

datYori:
we actually still use it in raw-utils

PR #108, opened 2023-08-22
RD-9445 Timestamp json output is different between truffle and scala

torcato:
Scala compiler uses just one format string with milliseconds .  So I made the same for truffle.

Quick question:
there was an specific if then else for the case of times/timestamps with milliseconds or not.
Do we prefer this?

Usually its better to have a fixed format but  the format string ` HH:mm:ss[.SSS]` would still parse both cases (with milliseconds and without).

If that is the case I can change the the scala compiler instead


Alexzerntev:
In truffle implementation we followed the Scala one, so I guess we missed it

PR #115, opened 2023-08-22
Remove old deprecated code.

miguelbranco80:
Also fix some naming conventions.


raw-compiler-rql2-truffle/src/main/scala/raw/compiler/rql2/truffle/Rql2TruffleCompiler.scala
```diff
@@ -272,7 +272,12 @@ class Rql2TruffleCompiler(implicit compilerContext: CompilerContext)
 
     val rootNode: RootNode = outputFormat match {
       case "csv" =>
-        val lineSeparator = if (programContext.settings.getBoolean(WINDOWS_LINE_ENDING)) "\r\n" else "\n"
+        val windowsLineEnding = programContext.runtimeContext.environment.options.get("windows-line-ending") match {
```
Alexzerntev, line 275: Should this string also be substituted with some const?

bgaidioz (approved):

PR #116, opened 2023-08-23
RD-9446: Sql.Read/InferAndRead implemented as sugar for .Query

bgaidioz:
Databases `.Read` calls are here implemented as sugar for `.Query`. This patch also impacts the scala executor which I have tested locally. I'll create a release from this branch in order to test Scala in CI.
* existing unimplemented `TruffleRead` nodes have been deleted. Desugaring takes care of the extension now.
* tests were re-enabled.

There are more changes than one would expect. When enabling more tests, some expected failures were exercised and our error message wasn't matching the expected one. The reason being that we didn't distinguish "column errors", hit when trying to decode a certain column as a certain type, from "database errors", hit for example on a connection failure.

Therefore I add a new exception and make sure it's triggered from all the failing column parsers. As part of that change, I renamed `idx` as `colName` since effectively the parameter is a `String` column name.


bgaidioz:
rerun tier1

bgaidioz:
run tier1 tests

bgaidioz:
run tier1

torcato (approved):
I see that you added the quotes as in the MysqlEmitter, PostgresqlEmitter, etc.

PR #120, opened 2023-08-24
RD-8898 Code formatter replaces triple quotes by single quotes

torcato:


torcato:
/ publish

bgaidioz (approved):
OK nice. I don't have a better idea. For a while I thought you had to implement a custom Truffle node and I was going to propose some desugaring from `TripleQuoteStringConst` to `StringConst` during a compiler phase since this is all for the early parser/pretty-printing. But OK that one didn't introduce many changes.

Also, I looked quickly to see if we were pattern-matching `_: StringConst` in random places where we should expect both cases but I didn't find any.

PR #121, opened 2023-08-24
RD-9462 propagation uncovered-case

Alexzerntev:
- Made Error.Get to expect tryable regardless if it is a nullable (provided by @bgaidioz )
- Added exception on getting a failure and getting an error on success


bgaidioz (commented):
I agree it's annoying everything gets reindented :)

raw-compiler-rql2/src/main/scala/raw/compiler/rql2/builtin/ErrorPackage.scala
```diff
@@ -74,7 +74,7 @@ class ErrorBuildWithTypeEntry extends EntryExtension {
       optionalArgs: Seq[(String, Arg)],
       varArgs: Seq[Arg]
   )(implicit programContext: ProgramContext): Either[String, Type] = {
-    val TypeArg(t) = mandatoryArgs(0)
```
miguelbranco80, line 77: I find this less readable. We don't think of arguments as being head and tail. We think of them as index positions. The suggestion to use .head is because we use a Seq to represent arguments. If it were an Array - probably should be - then it wouldn't even suggest this.

Alexzerntev: At some point we said to get rid of all the warnings, that is why I "correct" them.

miguelbranco80: Do you recall if this an Intellij 'best practices' notice, or an actual Scalac compiler warning?

Alexzerntev: Its Intellij "inspection" on scala code, a default one. But still shouldn't the new contributes which use intellij see no warning?

miguelbranco80: The focus was mostly Scala compiler warnings but these Intellij suggestions are often good. I did run a few days the 'Code Cleanup' task and it also suggested a variety of healthy changes, also to Scala code.
That said, here I think the code is just harder to understand. The underlying issue is the signature of the method so the Intellij suggestion is 'wrong' for our intended purposes. So I wouldn't apply this particular change.

Alexzerntev: Ok, changed for this PR. But since I was doing it for some time there are other files too. I will do it gradually if it is ok, or in a separate PR dedicated to that

bgaidioz (approved):

PR #145, opened 2023-08-28
style: javafmtAll

datYori:
In this PR we only apply the `javafmtAll` sbt task coming from #146 

- [x] #146


bgaidioz (approved):
I'm fine with 4 spaces :)

PR #152, opened 2023-08-29
Fixed RD-5910: XML typechecker shouldn't accept complex types for attributes

bgaidioz:
The XML typechecker shouldn't accept complex types for attributes. We can cope with `list` and `collection` because those are parsed item by item, and that would lead to lists and collections of one item. Why not. Like for fields. But records no.


torcato (approved):
LGTM

PR #153, opened 2023-08-29
Fixed RD-5986: Json.Parse/Read shouldn't successfully parse lists and objects are strings

bgaidioz:
The Truffle parser doesn't have that bug. Maybe the Scala one will need a fix if running that test.

I did add a couple of tests and a better error message (it was wrong: saying "scalar value found" when it shouldn't be "not found"). Anyway the error better states what wasn't expected?


torcato (approved):
LGTM

PR #154, opened 2023-08-29
RD-9409: Fix the flexible parser (LSP) to accept an extra comma before the `in`

bgaidioz:
The regular inflexible parser is OK with that comma. We should instead accept it in the flexible parser.


torcato (approved):
LGTM

PR #160, opened 2023-08-30
RD-9504 change GraalVM arguments for testing

Alexzerntev:
- Changed the CompileImmediately option and removed the dumping of graphs (otherwise the instance will be out of space at some point) 

Here you can find the truffle options : https://www.graalvm.org/22.0/graalvm-as-a-platform/language-implementation-framework/Options/

What I would like to discuss is 
- OSR, I think we should allow it because the iterators are basically large loops so OSR makes sence
- There are some options for TruffleBoundaries inclining. Should we consider them too? Maybe it improves jackson parsers performance.

@torcato @miguelbranco80 @bgaidioz


miguelbranco80:
@Alexzerntev (and @datYori ), these are the development settings and hence this isn't the right place to change settings. Rather, @datYori has to apply changes to the deployment templates.

datYori (approved):
the title is a bit confusing though sine we are changing only the java options used for the tests here

PR #166, opened 2023-09-01
RD-9485 Xml inferrer samples objects but this info is not passed upstream

torcato:


raw-compiler-rql2/src/main/scala/raw/compiler/rql2/builtin/XmlPackage.scala
```diff
@@ -92,13 +100,20 @@ class InferAndReadXmlEntry extends SugarEntryExtension with XmlEntryExtensionHel
...
+      val makeTryables = sampled
+
+      val t = inferTypeToRql2Type(dataType, makeNullables, makeTryables) match {
         case Rql2IterableType(inner, _) => Rql2IterableType(inner)
         case t => addProp(t, Rql2IsTryableTypeProperty())
       }
+      // validateXmlType changes the type, for instance, it removes the tryable property or type and attributes
+      // So we are running it here
+      validateXmlType(t).right.get
     }
```
torcato, line 117: I am calling validateXmlType, so that the types are consistent .
Should I create a new function inferTypeToXmlRql2Type ?

torcato:
retest this please

bgaidioz (approved):

PR #171, opened 2023-09-05
RD-9528: slow phases

bgaidioz:
The bottom commit "initial setup for benchmarking" is changing log settings, some logging message level, and _artificially multiplying phase rewrites with a loop_. Also it adds a test that takes long to run because of the code size. We can probably remove that commit eventually. It's there for testing.

The fixes are in the commit after. They mostly use `Any` in `query` and `everywhere`, wherever the profiler pointed us to. The left one creating `ClassCastException` is now `org.bitbucket.inkytonik.kiama.relation.Tree$$anonfun$bottomupNoBridges$1.applyOrElse(Object, Function1)`.


torcato (approved):
LGTM

raw-compiler-rql2/src/test/scala/raw/compiler/rql2/tests/regressions/RD9528Test.scala
```diff
@@ -0,0 +1,545 @@
...
+    |                if(Nullable.IsNull(organization))
+    |                then
+    |                if(include_internal_org) then true else Regex.Matches(c.org_name, "dev.*")
+    |                else c.org_name==(organization)
+    |        )
+    |
+    |""".stripMargin
+
+  test(v5_common_snapi + " 1 + 2") {
+    _ should (run and run and run)
```
bgaidioz, line 542: make a loop? With a parameter set to 1.

PR #182, opened 2023-09-11
RD-9556: Fix TRACE logLevel

bgaidioz:


raw-compiler/src/main/scala/raw/compiler/scala2/Scala2JvmCompiler.scala
```diff
@@ -155,13 +155,12 @@ class Scala2JvmCompiler(
       }
     })
 
-    logger.trace("Compiler classpath: " + compiler.classPath.asURLs.mkString("\n"))
```
bgaidioz, line 158: I moved it to _after_ because it's null.

raw-compiler/src/main/scala/raw/compiler/base/TreeWithPositions.scala
```diff
@@ -149,34 +149,36 @@ abstract class TreeWithPositions[N <: BaseNode: Manifest, P <: N: Manifest, E <:
...
-            startCol = positions.getStart(e).get.column
-            endCol = line.length
-          } else if (positions.getFinish(e).get.line == lineno) {
-            startCol = 0
-            endCol = positions.getFinish(e).get.column
-          } else if (lineno > positions.getStart(e).get.line && lineno < positions.getFinish(e).get.line) {
-            startCol = 0
-            endCol = line.length
-          }
+          for (start <- positions.getStart(e); finish <- positions.getFinish(e)) yield {
```
bgaidioz, line 152: Sometimes nodes have no position (the one I noticed was `Package("Record").Build`

raw-compiler/src/main/scala/raw/compiler/base/TreeWithPositions.scala
```diff
@@ -149,34 +149,36 @@ abstract class TreeWithPositions[N <: BaseNode: Manifest, P <: N: Manifest, E <:
           val lineno = _lineno + 1
           output += line + "\n"
 
-          var startCol = -1
-          var endCol = -1
-
-          if (positions.getStart(e).get.line == lineno && positions.getFinish(e).get.line == lineno) {
```
bgaidioz, line 155: Sometimes nodes have no position (the one I noticed was `Package("Record").Build`). The `.get` crashes (all this is part of the code that gets triggered by a `.trace` somewhere).

miguelbranco80 (approved):

PR #191, opened 2023-09-13
RD-9528: Monkeypatched Kiama files + improvements to transpilation time

bgaidioz:
The benchmark consists of 100 runs of a large piece of Snapi code, ending with simple expression `1+2`. This takes a long time to parse and transpile, and executes quickly since it's only computing `1+2`.

To measure the performance this is passed 100 times through `run` with `should run and run and ... and run` (100 times). Training-wheels are disabled, and GraalVM flags are modified to not dump the graph files to disk. The GraalVM changes is removed from the changeset.

| version | 100×`run` |
| ------- | --------- |
| main    |  11 minutes, 15 seconds       |
| patched  `Tree.scala` | 6 minutes, 32 seconds       |
| patched `Cloner.scala` | 4 minutes, 40 seconds      |
| more `rule[Any]` | 2 minutes, 32 seconds  |
| avoiding calls to `levenshteinDistance` | 1 minute, 32 seconds |


miguelbranco80:
So we go from 6.7s to 0.9s per query 👍

miguelbranco80 (approved):

PR #192, opened 2023-09-13
RD-9554: Fixed {Int/Long/Timestamp}.Range which couldn't be read multiple times

bgaidioz:
Added the tests all failing without fixes. Then fixed it. I couldn't find a shorter fix.

### DIFF
--- a/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/numeric/int_package/IntRangeNode.java
+++ b/raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/numeric/int_package/IntRangeNode.java
@@ -17,7 +17,7 @@ import com.oracle.truffle.api.dsl.Specialization;
 import com.oracle.truffle.api.nodes.NodeInfo;
 import raw.runtime.truffle.ExpressionNode;
 import raw.runtime.truffle.runtime.exceptions.RawTruffleRuntimeException;
-import raw.runtime.truffle.runtime.iterable.sources.RangeCollection;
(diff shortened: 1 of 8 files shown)


raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/numeric/int_package/IntRangeNode.java
```diff
@@ -29,6 +29,6 @@ public abstract class IntRangeNode extends ExpressionNode {
   protected Object doRange(int start, int end, int step) {
     if (step <= 0)
       throw new RawTruffleRuntimeException("range step has to be strictly positive", this);
-    return new RangeCollection(start, end, step);
```
miguelbranco80, line 32: Feel free to ignore as not familiar with this but in case: should we be removing some unused RangeCollection now?

bgaidioz: I don't understand the question. What is unused?

miguelbranco80 (approved):

raw-runtime-rql2-truffle/src/main/java/raw/runtime/truffle/ast/expressions/builtin/numeric/int_package/IntRangeNode.java
miguelbranco80: Sorry, you can ignore. I missed the file rename and got confused.

PR #195, opened 2023-09-14
Fixed RD-8890: com.amazonaws.AbortedException

bgaidioz:


raw-sources-s3/src/main/scala/raw/sources/filesystem/s3/RawS3InputStream.java
```diff
@@ -0,0 +1,39 @@
...
+
+package raw.sources.filesystem.s3;
+
+import com.amazonaws.services.s3.model.S3ObjectInputStream;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InterruptedIOException;
+
+public class RawS3InputStream extends InputStream {
```
miguelbranco80, line 21: Sorry, this is a Java file in a scala source folder. What's the reason?

bgaidioz: I had two implementations while experimenting with exception handling, and didn't pay attention I kept the Java one.

raw-sources-s3/src/main/scala/raw/sources/filesystem/s3/RawS3InputStream.scala
```diff
@@ -0,0 +1,35 @@
...
+package raw.sources.filesystem.s3
+
+import com.amazonaws.AbortedException
+import com.amazonaws.services.s3.model.S3ObjectInputStream
+
+import java.io.IOException
+import java.io.InputStream
+import java.io.InterruptedIOException
+
+class RawS3InputStream(private var s3ObjectInputStream: S3ObjectInputStream) extends InputStream {
```
miguelbranco80, line 22: Should be a private val I think.

miguelbranco80 (approved):

raw-sources-s3/src/main/scala/raw/sources/filesystem/s3/RawS3InputStream.scala
bgaidioz: I relied on the Java to Scala translation and didn't see that. I think it's OK without `private val`.

PR #206, opened 2023-09-21
RD-9624 wrap byte array and big decimal into truffle object

Alexzerntev:
- Wrapped `byte[]` and `BigDecimal` in truffle objects
- Fixed jdbc assertions
- Fixed XML parser assertions


Alexzerntev:
Merged in https://github.com/raw-labs/snapi/pull/220

PR #207, opened 2023-09-22
Support Graal 21 + Modularization of language

miguelbranco80:
This (massive) PR is the first step towards supporting Graal 21 and language modularization.
Motivation: Graal 21 was released this week. There are various significant changes/improvements.
The most notable one - pointed out by Christian Humer in Slack, blog posts, etc - concerns the modularization needs.
Essentially, we now need to modularize the language and split it fully from the launcher.
Without this, there is no guarantee we get performance or important optimization phases even run.

What is modularization?

Let’s get ChatGPT’s help:

“””
Java 9 introduced a significant new feature called the Java Platform Module System (JPMS), often referred to simply as "modularization." Before Java 9, applications and libraries were packaged primarily as JAR files with no intrinsic notion of which other JARs they required or what parts of them were meant for public consumption versus internal use. This lack of a defined structure could lead to issues such as classpath hell, where multiple versions of a library could conflict, or unintentional use of internal APIs.
Modularization in Java 9 was introduced to address these issues and more. Here are the primary aspects of JPMS:
Modules: A module is a named, self-describing collection of code and data. It can export specific packages for use by other modules and declare its dependencies on other modules.
* module-info.java: This is the descriptor for a module. It describes which packages the module exports, which other modules it requires, and other module-related configurations.
* Exports and Requires: In the module-info.java, a module declares which packages it exports (makes available to other modules) with the exports directive. It also states which other modules it depends on with the requires directive.
* Encapsulation: By default, a module's internal implementation details (those packages not explicitly exported) are hidden from other modules, leading to stronger encapsulation.
* Migration: Recognizing that not all applications and libraries would immediately become modularized, JPMS provided a way to gradually migrate code from the classpath to modules. The key feature aiding in this is the "unnamed module," where all classpath-based code resides. There are also "automatic modules" which allow JARs to be placed on the module path and treated as modules without a module-info.java descriptor.

Some challenges arose with the introduction of JPMS. Many existing libraries and frameworks had to adapt to this new system, and not all saw it as an improvement. Some projects decided to hold off on adopting modules, while others embraced it. Either way, JPMS is a significant shift in the Java ecosystem, encouraging developers to think more about explicit dependencies and the structure of their code.
“””

Why modularization for Graal?

Well, it makes sense. It’s akin to the Truffle boundary in a way. If a language is a module that exports *nothing* - except RawLanguage - then Graal/JIT/Compiler is sure what it can optimize & what for sure cannot ever change. It provides the ultimate barrier.
This does mean we really need to split the “launcher” from the “language”. The language is single JAR module, and the launcher is a separate one.

What does it mean to us?

Quite a lot actually.
Here’s the structure I ended up with.
This PR rebuilds the repo, essentially from scratch.
The directory structure is now:
```
language/
extensions/
deps/
	kiama/
	scala-logging/
launcher/
```

Let’s go from them in turns.

We start by deps/
With module systems, we cannot monkey patch Kiama anymore. That’s exactly what modules forbid.
deps/ contains Scala packages which we need modularized, and which are not.
Essentially, for now I copied those projects, applied changes to modularize them.
We will of course contribute those changes to the repos themselves but this was easiest to get started.
Each deps/ has a README-RAW-LABS.md describing the changes done.
But essentially, the changes are: make the package a Java module; and in the case of Kiama, apply the monkey patches in the source files.
What does it mean to make a package a module? Well, there’s two ways: one is to really make it a module - add a module-info.java file specifying the module dependencies and exports. The other is well, do nothing. If we pick ANY JAR and add it to the modulepath - instead of the classpath - the JDK will automatically turn it into an ‘automated module’. Since there is no module-info.java, the JDK will use the JAR name as the module name. And it exports everything in the package.
So you may wonder why do I have ‘scala-logging’ in our deps. Can’t that just be handled by the JDK and be turned into an automatic module? Sadly no. Scala complicates things. When the JDK picks up a JAR name to turn it into the module name, it strips out the version number. So “google-guava-2.10.jar” becomes module “google.guava”. If I then need it in a project I need to add “requires google.guava” to my module-info.java. But Scala screws this up. Because Scala names the JARs as “scala-logging_2.12-0.5.jar”. Notice the _2.12 which is the Scala compiler version. So this would become a module name “scala.logging.2.12”. This is in invalid module name! It’s impossible to do “requires scala.logging.2.12” because numbers are not allowed as identifiers!
We aren’t the only ones suffering from this :) In fact, the JDK has a system where we can impose a module name of a JAR that hasn’t been modularized yet. This consists in adding a special string (Automated-Module-Name) to the JAR file.
So this is what I had to do for scala-logging. I changed the build.sbt to add the Automated-Module-Name of “scala.logging” so that I could “requires” it from the RAW language module. Argh.
We will need to do this for *every Scala dependency*. Fortunately we have very very few true Scala dependencies (only 4 in total). For instance, the Scala compiler is a dependency but they already include a good Automated-Module-Name in their JAR.
So deps, in summary is there for Kiama and for scala-logging fixes.

What about language vs extensions?

Well, the language now has to be a JAR module. The language/ directory includes (almost) all the RAW code. It includes the Scala code for the compiler packages, sources, utils, inferrer, and base runtime packages. It also includes the Java runtime code. To build a JAR module, Java has to run last. So we set the compilation order to compile first the Scala code, then the Java code. This makes sense - the Scala code is the transpiler, pretty printer, parser, etc - and the Java code is the runtime. All good here. It’s a module and if you look at language/src/java/module-info.java you see what it requires, etc.
But then we have an issue. We also have Scala code that depends on Java code. The TruffleCsvPackage and friends is Scala code that depends on Java code. Well, if Java runs after Scala, then I cannot run Scala again :-) So the solution I came up with was to split this Scala code - that depends on Java code like nodes - onto a separate project altogether. This is the extensions/ project.
This is horrific to use and a pain.
We need to convert all those extensions - and Rql2TruffleCompiler “doEmit” - to Java code. At that point we have a logical split - Scala for kiama stuff; Java for everything else, all in the language project. But alas, we cannot do that now.
That’s why “extensions/“ exists.
This is a pain as we need version numbers - build language, then build extensions, etc. We use the same tricks as in the main repo - there’s a version.properties file controlling these version numbers.

Finally, we have the launcher.
The launcher HAS to be a Maven Java project now. (Scala and SBT just aren’t aware from the module system.) This launcher project depends on launcher and on extensions and loads both of them as modules. The launcher should become the new “raw CLI”. For now, I made a simpler demo in there. The old (Scala-based) CLI code is still there but needs to be converted to Java. But now it’s truly Polyglot. The launcher I have there even mixes Python and Snapi - this all works now.

So, where do things stand?
Considering the magnitude of the change, pretty OK I’d say.
First off, the old Scala code generator works fine. There’s a PR against the main repo - https://github.com/raw-labs/raw/pull/4850 - that applies the necessary changes. Basically, the main change is that we only have a single JAR for RAW now.
The other change is due to an API change. Turns out that if we have a ServiceLoader defined in raw.creds, and an instance defined in raw.creds.local or raw.creds.jdbc, this is disallowed by the module system. Module system forbids service loaders in nested packages. So I had to move raw.creds to raw.creds.api, so we now have raw.creds.api defining the interface and ServiceLoader, and raw.creds.local and raw.creds.jdbc defining an implementation. Makes sense actually - otherwise a nested package could overwrite stuff, a bit like monkey patching.
I only did this change to raw.creds but I’ll need to do it in others like raw.auth, etc.
The thing is that with the module system things don’t fail at compile time. They fail at runtime, when it tries to load the class. Since the only Truffle tests I run are trivial, I never needed or touched other problematic packages.

Truffle tests - basic ones - also run. Now, in Graal 21 - which you need (sdk install java graal-21), the option names are different. Namely, locator.disabled is gone. Refer to extensions/build.sbt for the new names. The great thing is that all those add-opens and add-exports aren’t needed anymore. But remember to update your Intellij Run configuration settings!
I expect MOST Truffle tests to fail. There’s more module-info.java to open up with “requires” (refer to language/src/java/module-info.java) for sure.

Also, I broke the “emitMethods”. emitMethods was doing RawLanguage.getCurrentContext to access a FunctionRegistry. This is a violation because it was at compile time - AST generation time - trying to add “runtime stuff”. This is not allowed anymore. It just isn’t correct anyway to do that. So we will need to redo emitMethod to, say, add functions to a FrameDescription and reference them. (BTW, there was a similar “abuse” of FunctionRegistry for closure nodes, but I’m told by Alex Z. that closures don’t need FunctionRegistry so I just deleted that). This was not me messing around. This is stuff that prevented any Truffle test from even starting.

Another one is the context and stage compiler. We used to create a context during emission and closing it at the end of the writeTo method. In the doEval case - staged compiler codepath - we did something similar. This is just wrong - we knew that. Only the launcher creates the context. These modules cannot have anything to do with it.
For test cases, however, I made the “doExecute” method of the test suite create a context and shut it down. Of course this only works with -Dpolyglotimpl.DisableClassPathIsolation=true (that’s the “new variant” of the old locator disabled flag)

Some random caveats:
- Kiama for some reason didn’t build on Graal 21. Needs a “normal” JDK. I have a build.sh script committed showing a sample version that works fine.
- Doc generation/publication was disabled. The doc generator got confused when it saw module-info.java as it thought it was a Java file - it’s not really, has a special syntax. It’s one of those things that shows SBT and company are just not ready for JAva modules
- Because the launcher is in Maven, I changed all build.sbt scripts (language/, extensions/, kiama/, scala-logging/) to also publish to the local Maven repo. SBT uses Ivy. Maven uses M2. They are different layouts/directories, so I made sbt publishLocal send JARs to both so that both SBT and Maven find e.g. Kiama or what not.

So, what’s left to do:
- we need to find a way to manage Kiama, scala-logging (and maybe other dependencies in the future).
- fix the CI to build this thing properly. Once it publishes the language/ (and kiama and scala-logging which the compiler needs), then the other PR in the internal repo will be ready to go as to support the legacy code generator. Note that extensions/ or launcher/ aren't needed for the legacy code generator, only the language/
- we need to get rid of extensions/ by converting *everything* in there to Java (!!!) and then just moving it to language/src/java/
- ~~build.sbt are a mess. I will clean them up. I have comments there to remind me of ideas even so don’t clean them.  Please ignore; I wIll fix.~~
- ~~I didn’t test the doEval codepath. Might be broken?~~
- the emitMethods has to be reimplemented
- docs generation is broken (not that we care but I thing the open source repo we publish to cares?)
- we need to remove RawLanguage/RawContext methods related to runtime context and outputstream. One implication of this is that we need to handle the output stream by rewriting the *source code* to say "write_csv" and receiving "outputstream" as an argument (or some similar technique)
- I did a bunch of API hacks so that extensions/ and language/ see each other. These are to clean up once extensions/ is gone. One example is rawLanguageAsAny in the doEmit. Please ignore; I wIll fix.
- need to find a way to delete RawLanguage.getCurrentContext method - and hence, all its uses - since that method is just wrong…
- the launcher dependency versions (of language/ and extensions/) are hardcoded in launcher/pom.xml. I couldn't get Maven to read versions dynamically from a file; I don't think it's possible.

And later:
- we need to change the Runner to use Truffle contexts as well when running a Truffle execution!.

Now, if you want to see a cool demo, you must run raw.cli in the launcher project. For this:
- assuming Graal 21 is installed and set.
- update version.properties file
- cd deps/kiama/
- build.sh (which does publish local with a proper JVM which u need to have)
- cd deps/scala-logging/
- sbt publishLocal
- cd language/
- sbt publishLocal
- cd ../extensions/
- sbt publishLocal
- Then open launcher/ in Intellij, import as Maven project (point it to the pom.xml), make sure it’s using Graal 21, and run the RawCli.


miguelbranco80:
I've now:
- cleaned the sbt files of language/ and extensions/
- fixed the doc generation to skip all Java files (it was not just module-info.java but all Java files causing issues; I guess the doc generator just doesn't understand Java code)

I've also confirmed staged compiler tests fail. I think it's a funny one - the validation (typing) executes code and that needs a context now. So I think I need to have the validation method of the test suite itself creating a Truffle context...

miguelbranco80:
> I've also confirmed staged compiler tests fail. I think it's a funny one - the validation (typing) executes code and that needs a context now. So I think I need to have the validation method of the test suite itself creating a Truffle context...

I've confirmed this is the case and fixed it. Basically, in the TruffleCompilerTestContext I override methods (like doValidate) by wrapping them to create/close the TruffleContext.

PR #211, opened 2023-09-26
[Graal21] RD-9639 emitMethods + other fixes.

bgaidioz:
Let's review the individual commits one by one.
* I fixed `emitMethods` by letting the top program evaluate to an `ExpBlock`.
* Datasets weren't loading because the hack to get them extracted as files was filtering the module names with an old name.
* `Location.Describe` had been patched to return the `type` as an `Rql2Type` instead of a `String`. I suppose it was done as part of migrating and trying to compile. But I tried to reset it to use the `PrettyPrinter` and it passed.
* `Snowflake` isn't working for me, there's a failure when decoding rows using Arrow. I tried a number of options that would help a bit without fixing all problems. The working one was found in a [Snowflake Github issue](https://github.com/snowflakedb/snowflake-jdbc/issues/589) which says there's something not working in Apache Arrow with new JVMs, and a person suggested to get results as a JSON. This changes slightly the error messages though.

With these patches I think all tests pass. The only failures I got were due to the flaky `fail-after-10` tests. When running them one by one they would pass.


language/src/test/module-info.java
```diff
@@ -0,0 +1,147 @@
+module raw.language {
```
miguelbranco80, line 1: In my new commit to to the PR yesterday, I added a lot of missing module-info.java entries. Apologies for neglecting to mention that. That said, I added those to src/main/module-info.java. You may want to try with that instead. Perhaps merge yours with mine (and place it in module-info.java) because I may still have neglected some.
I believe that src/main/module-info.java is the right place to add those as I believe it makes it visible to tests as well. For sure we need it in 'main' for the launcher to work: the launcher needs these open for the system to work. Perhaps by adding them to 'test' is enough for the extensions to work, since the extensions only runs as a testsuite right now.
Give it a try with a bigger correct src/main/module-info.java (as per my last commit) and then we see. Or maybe there's something else I don't understand.
But otherwise, looks good!! If that all works feel free to merge to my own PR even, so that it's ready to pass the CI.

PR #212, opened 2023-09-26
RD-9598: Missing error case in Errors.PrettyPrinter

bgaidioz:
Added a test to trigger the bug and the fix.

I also walked through all errors and ensured they're referenced in `ErrorsPrettyPrinter`.


torcato (approved):
LGTM

PR #213, opened 2023-09-26
RD-9616: Fix the computation of Interval.ToMillis

bgaidioz:
Rounding is done last so that we don't round days. It was a mistake. Added a test that would fail without the patch (both computations would mistakenly return 30 days).


torcato (approved):
LGTM

PR #220, opened 2023-09-28
Fixing assertions new structure

Alexzerntev:


language/src/main/java/raw/runtime/truffle/runtime/primitives/DecimalObject.java
```diff
@@ -0,0 +1,28 @@
...
+
+package raw.runtime.truffle.runtime.primitives;
+
+import com.oracle.truffle.api.interop.TruffleObject;
+import java.math.BigDecimal;
+
+public class DecimalObject implements TruffleObject {
+  private final BigDecimal bigDecimal;
+
+  public DecimalObject(BigDecimal byteArray) {
```
bgaidioz, line 21: Maybe that parameter deserves a more suited name (like `bigDecimal`)? I suppose a copy/paste error.

PR #263, opened 2023-11-13
RD-9983 plugin antl4 parser

Alexzerntev:
- Switched Truffle compiler to antlr4 parser


snapi-frontend/src/main/scala/raw/compiler/rql2/antlr4/RawSnapiVisitor.scala
```diff
@@ -1014,52 +1298,79 @@ class RawSnapiVisitor(positions: Positions, private val source: Source, isFronte
...
+  }
+  override def visitNullableTryableType(ctx: SnapiParser.NullableTryableTypeContext): SourceNode = {
+    if (isFrontend) {
+      notFrontendError(ctx)
+    } else {
+      Option(ctx)
+        .map { context =>
+          val tipe = Option(context.tipe)
+            .flatMap(tipeContext => Option(visit(tipeContext)))
+            .getOrElse(ErrorType())
```
bgaidioz, line 1326: `ErrorType` isn't an `Rql2TypeWithProperties`, I suppose the `asInstanceOf` would fail if a type doesn't parse?

Maybe if the type doesn't parse, the properties should be parsed but ignored?

snapi-client/src/main/scala/raw/client/rql2/truffle/Rql2TruffleCompilerService.scala
```diff
@@ -78,25 +79,23 @@ class Rql2TruffleCompilerService(maybeClassLoader: Option[ClassLoader] = None)(
...
-    } catch {
-      case ex: CompilerParserException => ParseFailure(ex.getMessage, ex.position)
-      case NonFatal(t) => throw new CompilerServiceException(t, programContext.dumpDebugInfo)
+    val positions = new Positions()
+    val parser = new Antlr4SyntaxAnalyzer(positions, true)
+    val parseResult = parser.parse(source)
+    if (parseResult.isSuccess) {
+      ParseSuccess(parseResult.tree)
+    } else {
+      ParseFailure(parseResult.errors)
```
miguelbranco80, line 98: We need to handle NonFatal exceptions to get JIRA reports, as was done before.
Means we need to build the program context so we have everything we need to report in the JIRA crash report.

snapi-client/src/main/scala/raw/client/rql2/truffle/Rql2TruffleCompilerService.scala
```diff
@@ -108,17 +107,11 @@ class Rql2TruffleCompilerService(maybeClassLoader: Option[ClassLoader] = None)(
...
-          }
-        } catch {
-          case ex: CompilerParserException =>
-            GetTypeFailure(List(ErrorMessage(ex.getMessage, List(raw.client.api.ErrorRange(ex.position, ex.position)))))
-          case NonFatal(t) => throw new CompilerServiceException(t, programContext.dumpDebugInfo)
+        val tree = new TreeWithPositions(source, ensureTree = false, frontend = true)(programContext)
+        if (tree.valid) {
+          GetTypeSuccess(tree.rootType)
+        } else {
+          GetTypeFailure(tree.errors)
```
miguelbranco80, line 114: Two observations:
* We need to handle NonFatal exceptions to get JIRA reports, just as was done before.
* How are errors handled now? Can we successfully build a TreeWithPositions object with a broken tree? That is a significant API change which means we would need to always check the validity of the parse tree after building it. In the past, for safety, it would blow up on building.

snapi-client/src/main/scala/raw/client/rql2/truffle/Rql2TruffleCompilerService.scala
```diff
@@ -132,53 +125,44 @@ class Rql2TruffleCompilerService(maybeClassLoader: Option[ClassLoader] = None)(
...
+              rql2TypeToRawType(t) match {
+                case Some(rawType) => rawType
+                case None => return GetProgramDescriptionFailure(List(ErrorMessage("unsupported type", List.empty)))
+              }
+            },
+            comment
+          )
+          GetProgramDescriptionSuccess(programDescription)
+        } else {
+          GetProgramDescriptionFailure(tree.errors)
```
miguelbranco80, line 165: Need to handle NonFatal

snapi-client/src/main/scala/raw/client/rql2/truffle/Rql2TruffleCompilerService.scala
```diff
@@ -585,43 +569,40 @@ class Rql2TruffleCompilerService(maybeClassLoader: Option[ClassLoader] = None)(
...
+            case _: PackageNotFound => true
+            case _: NamedParameterAfterOptionalParameter => true
+            case _: ExpectedTypeButGotExpression => true
+            case _ => false
           }
-        } catch {
-          case NonFatal(t) => throw new CompilerServiceException(t, programContext.dumpDebugInfo)
+          ValidateResponse(formatErrors(selection, positions))
+        } else {
+          ValidateResponse(parseResult.errors)
```
miguelbranco80, line 605: Same issue with NonFatal.

snapi-client/src/main/scala/raw/client/rql2/truffle/Rql2TruffleCompilerService.scala
miguelbranco80: And if we are to change the API - which I'm not against but it will be time-consuming to get right I believe - we need a full review because I'm fairly confident we have many situations where we build the TreeWithPositions and wouldn't be checking that it parsed. Because in the past, it would just blow up and not even build the object.

And if we were to do that, we'd then have to review the existence of the CompilerParserException itself.

Or otherwise, we change CompilerParserException to carry multiple errors and start throwing as we used to do before, in the base TreeWithPositions.

miguelbranco80: For instance, look at this inside RawLanguage:

```
} catch (RuntimeException e) {
  if (e.getCause() instanceof CompilerParserException) {
    CompilerParserException ex = (CompilerParserException) e.getCause();
```

This would need to be adapted.

But it's not just places where CompilerParserException is used. After all, removing it would force you to find cases where it's leftover. But there's also CompilerException. Notice that CompilerParserException is a child of CompilerException so we'd also need to analyze every case where CompilerException happens and understand if we need to add the extra 'parse check OK' is needed or not, and propagate errors in the "consumer".

miguelbranco80 (changes requested):
There are issues with the CompilerParserException which we'd need to reconsider!

miguelbranco80:
It seems that if we were to delete CompilerParserException and CompilerExecutionException.
In the case of CompilerExecutionException we would replace it by throw CompilerException.
In the case of CompilerParserException, it's case-by-case basis: either replace by checking that parse errors are empty, or do as you're doing in the PR...

Alexzerntev:
Branch to be deleted. The branch that will be merged is : https://github.com/raw-labs/snapi/pull/317

PR #268, opened 2023-11-15
Fixes after exercising the internal parser against the internal pretty printer

bgaidioz:


snapi-frontend/src/main/scala/raw/compiler/rql2/antlr4/RawSnapiVisitor.scala
```diff
@@ -1136,7 +1136,6 @@ class RawSnapiVisitor(
             .getOrElse(ErrorType().asInstanceOf[Rql2TypeWithProperties])
           // this is needed for the case of parenthesis around nullable_tryable rule
```
bgaidioz, line 1137: remove

Alexzerntev (approved):

PR #274, opened 2023-11-21
Making antlr4 parser a separate package

Alexzerntev:
Moved antlr4 parser to separate package


miguelbranco80:
Q: Didn't we add the truffle dep to snapi-frontend because of antlr4?
If so, let's also remove it from snapi-frontend not to have it twice.

Suggest to call it snapi-parser. Also suggest to use the same style as the others for the build.sbt project name (raw-snapi-parser). This leaves space for other parsers we may need later, which may also be based on antlr4.

reset-versions.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+
+set -e
+
+
+cp /home/ld/workspace/raw-labs/snapi/utils/version /home/ld/workspace/raw-labs/raw/deps/utils.version
```
miguelbranco80, line 7: Please remove this script

snapi-frontend/build.sbt
```diff
@@ -160,6 +160,7 @@ libraryDependencies ++= Seq(
   snowflakeDeps,
   commonsCodec,
   springCore,
-  kryo
+  kryo,
+  rawSnapiParser
```
miguelbranco80, line 164: Move it to the top next to the other rawClient dependency. Suggest you add compile->compile;test->test in case we ever add test helpers.

snapi-frontend/project/Dependencies.scala
```diff
@@ -11,6 +11,9 @@ object Dependencies {
   val scalaLoggingVersion = IO.read(new File("../deps/scala-logging/version")).trim
   val kiamaVersion = IO.read(new File("../deps/kiama/version")).trim
 
+  val rawSnapiParserVersion = IO.read(new File("../snapi-parser/version")).trim
+  val rawSnapiParser = "com.raw-labs" %% "raw-snapi-parser" % rawSnapiParserVersion
+
```
miguelbranco80, line 16: Suggest you move it up closer to the other raw dependencies.

snapi-parser/build.sbt
```diff
@@ -0,0 +1,176 @@
...
+// Doc generation breaks with Java files
+Compile / doc / sources := {
+  (Compile / doc / sources).value.filterNot(_.getName.endsWith(".java"))
+}
+Test / doc / sources := {
+  (Compile / doc / sources).value.filterNot(_.getName.endsWith(".java"))
+}
+
+// Skipping javadoc generation for antlr4 broken links
+Compile / doc := { file("/dev/null") } // for Unix-like systems
```
miguelbranco80, line 76: BTW, do we still have this line in snapi-frontend build.sbt ? If so, remove it.

snapi-parser/src/main/resources/antlr4/antlr-4.12.0-complete.jar
miguelbranco80, line 1: This cannot be in resources because resources are data available at *runtime*, which is not the intention here. Place it in some other folder under src/

snapi-parser/src/main/resources/antlr4/SnapiParser.g4
miguelbranco80, line 1: This cannot be in resources because resources are data available at runtime, which is not the intention here. Place it in some other folder under src/

snapi-parser/src/main/resources/antlr4/SnapiLexer.g4
miguelbranco80, line 1: This cannot be in resources because resources are data available at runtime, which is not the intention here. Place it in some other folder under src/

snapi-parser/src/main/resources/antlr4/readme.md
miguelbranco80, line 1: This cannot be in resources because resources are data available at runtime, which is not the intention here. Place it in some other folder under src/

snapi-parser/src/main/resources/reference.conf
miguelbranco80, line 1: If there are no settings, this file is not necessary and should be removed.

miguelbranco80 (commented):
Some quick comments. Feel free to merge once these are applied!

PR #277, opened 2023-11-21
adapt ci for snapi-parser

datYori:


Alexzerntev (approved):
I will fix the errors that emerged

PR #279, opened 2023-11-22
Relax memory pressure when allocating inferrers

bgaidioz:
Declared all the formatters in an object. When profiling, those were dominating the memory. That fix permitted to run 78K semantic analysis without hitting OutOfMemory.

This would lead to a leak only if we never free Inferrers. Maybe this is the case?

### DIFF
--- a/snapi-frontend/src/main/scala/raw/inferrer/api/DateTimeFormatFinder.scala
+++ b/snapi-frontend/src/main/scala/raw/inferrer/api/DateTimeFormatFinder.scala
@@ -131,31 +131,3 @@ object DateTimeFormatFinder {
   }
 
 }
-
-trait DatetimeFormatFinder {
-
-  def getDate(str: String): Option[(String, LocalDate)] = {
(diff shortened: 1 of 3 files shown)


snapi-frontend/src/main/scala/raw/inferrer/api/DateTimeFormatFinder.scala
```diff
@@ -132,3 +131,31 @@ trait DatetimeFormatFinder {
   }
 
 }
+
+trait DatetimeFormatFinder {
```
miguelbranco80, line 135: Suggest you remove this and just import the object as a wildcard from the caller points. The trait is now unnecessary and someone could be tempted to accidentally add state to it.

bgaidioz: Like that?

miguelbranco80 (approved):

PR #280, opened 2023-11-22
Close RawContext and its inner services correctly.

miguelbranco80:


bgaidioz (approved):
After running 78K validations, I don't see a leak of `DateTimeFormatter` like we would see otherwise.

I see a bunch are allocated because we keep recreating new ones. And lots of string memory used because of the logTree function, but these aren't leaks, and are fixed by another PR.

PR #284, opened 2023-11-28
Fixed RD-10197: Too many allocations of RawSettings + CredentialService API

bgaidioz:


snapi-truffle/src/main/java/raw/runtime/truffle/RawLanguage.java
```diff
@@ -64,6 +67,11 @@ public final class RawLanguage extends TruffleLanguage<RawContext> {
   public static final String VERSION = "0.10";
   public static final String MIME_TYPE = "application/x-rql";
 
+  public final RawSettings rawSettings =
+      new RawSettings(ConfigFactory.load(), ConfigFactory.empty());
+  CredentialsService credentialsService =
```
miguelbranco80, line 72: Shouldn't it be public final?
If it works I'm ok with this. I thought you had to do it in some "sort of constructor" but this works.
You need to test it in full JPMS mode. Make a release, create a Docker, and test if credentials work.
It either works or not.

bgaidioz: Yes, I forgot the `public final`. And OK I try with a docker.

bgaidioz: I booted a docker compose with both the runner and creds. With the config to make creds visible to the runner.

Is it OK if `Environment.Secret("tralala")` properly evaluates to `could not find secret tralala`? That means it went through the whole java module loading I think.

miguelbranco80 (approved):

PR #288, opened 2023-12-01
RD-10206: Setting RDBMS fetch size from config

bgaidioz:
When running `PostgreSQLPackageTruffleTest`, that line appears in the logs when using a `PostgreSQL` database. Same with other backends.
```
Using raw.runtime.rdbms.fetch-size: 100000 (int)
```

Some tests processing errors using the `Exception` out of the JDBC call, failed because of missing module declarations.
```
Actual: java.lang.IllegalAccessError: class raw.sources.jdbc.pgsql.PostgresqlClient (in module raw.snapi.frontend) cannot access class org.postgresql.util.PSQLException (in unnamed module @0x479d31f3) because module raw.snapi.frontend does not read unnamed module @0x479d31f3
```

### DIFF
--- a/snapi-frontend/src/main/java/module-info.java
+++ b/snapi-frontend/src/main/java/module-info.java
@@ -21,6 +21,10 @@ module raw.snapi.frontend {
   requires java.sql;
   requires com.ctc.wstx;
   requires kiama;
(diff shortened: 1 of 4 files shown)


torcato (commented):
LGTM

PR #289, opened 2023-12-03
Truffle optimizations 1

Alexzerntev:
## The first iteration of truffle optimizations

- Introduces `ErrorObject()` which is equivalent of Tryable.Failure. 
- And `NullObject` which is a singleton. 

All the places where nullable - tryables are used are now either a *value* of one of the above objects. Correspondent static  methods are created to check if the value is Success or NotNull.


bgaidioz (approved):
Only minor comments.

PR #290, opened 2023-12-04
RD-9983 + RD-9483 + RD-8687 all lsp changes

Alexzerntev:
All the latest lsp changes from the tittle PRs


bgaidioz (commented):
The `Rql2TruffleCompilerService.validate` method hasn't been patched. The thing is, since it's using `withLspTree`, it shouldn't need to recreate a new `Tree` to report syntax errors. That is needed in `main` as there are two parsers (the flexible and the inflexible one), but now the parser is the flexible one, and `validate` should basically only report the errors it found.

Probably something to simplify here, to avoid reparsing.

Alexzerntev:
Branch to be deleted. The branch that will be merged is : https://github.com/raw-labs/snapi/pull/317

PR #293, opened 2023-12-06
RD-10218: Don't report internal errors (exceptions due to internal errors) as a user-visible error

bgaidioz:
If hitting an internal error, the exception is caught and mistaken for a user-level one. The discrimination on `.isGuestException` isn't enough.

Our language can throw:
* `RawTruffleRuntimeException` and `RawTruffleValidationException` that have a user-visible message and can be reported as an execution or validation error to the user,
* `RawTruffleInternalErrorException` which is showing an internal error,
* any other regular exception that it triggers by calling libraries: `NoSuchElementException`, `NullPointerException`, etc.

All these exceptions end up wrapped as `PolyglotException`. I noticed method `.isInternalError` returns `true` for regular Java exceptions and `false` for `RawTruffleRuntimeException` and `RawTruffleValidationException`. By checking `.isInternalError`, we get the intended behavior: `RawTruffleRuntimeException` is caught and returned as `ExecutionRuntimeFailure`, and regular unexpected ones are propagated as a `CompilerServiceException`

I believe our `RawTruffleInternalErrorException` should be recognized as internal. Because it inherits `AbstractTruffleException`, it's seen as non-internal and is confused with a `RawTruffleRuntimeException`. So far the only way I was able to distinguish it, was to drop the inheritance with `AbstractTruffleException`.

### DIFF
--- a/snapi-client/src/main/scala/raw/client/rql2/truffle/Rql2TruffleCompilerService.scala
+++ b/snapi-client/src/main/scala/raw/client/rql2/truffle/Rql2TruffleCompilerService.scala
@@ -404,11 +404,8 @@ class Rql2TruffleCompilerService(maybeClassLoader: Option[ClassLoader] = None)(
             }
             ExecutionValidationFailure(errors.to)
           } else {
-            if (ex.isInternalError) {
-              val programContext = getProgramContext(environment.user, environment)
-              throw new CompilerServiceException(ex, programContext.dumpDebugInfo)
-            }
-            ExecutionRuntimeFailure(ex.getMessage)
+            val programContext = getProgramContext(environment.user, environment)
+            throw new CompilerServiceException(ex, programContext.dumpDebugInfo)
           }
         } else {
           // Unexpected error. For now we throw the PolyglotException.
--- a/snapi-client/src/main/scala/raw/client/rql2/truffle/Rql2TruffleCompilerService.scala
+++ b/snapi-client/src/main/scala/raw/client/rql2/truffle/Rql2TruffleCompilerService.scala
@@ -404,6 +404,10 @@ class Rql2TruffleCompilerService(maybeClassLoader: Option[ClassLoader] = None)(
             }
             ExecutionValidationFailure(errors.to)
           } else {
+            if (ex.isInternalError) {
+              val programContext = getProgramContext(environment.user, environment)
+              throw new CompilerServiceException(ex, programContext.dumpDebugInfo)
+            }
             ExecutionRuntimeFailure(ex.getMessage)
           }
         } else {
--- a/snapi-client/src/main/scala/raw/client/rql2/truffle/Rql2TruffleCompilerService.scala
+++ b/snapi-client/src/main/scala/raw/client/rql2/truffle/Rql2TruffleCompilerService.scala
@@ -412,7 +412,9 @@ class Rql2TruffleCompilerService(maybeClassLoader: Option[ClassLoader] = None)(
               }
               ExecutionValidationFailure(errors.to)
             } else {
-              // A runtime failure during execution. The query is a failed tryable.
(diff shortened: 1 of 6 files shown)


snapi-client/src/main/scala/raw/client/rql2/truffle/Rql2TruffleCompilerService.scala
```diff
@@ -404,7 +404,12 @@ class Rql2TruffleCompilerService(maybeClassLoader: Option[ClassLoader] = None)(
...
-            ExecutionRuntimeFailure(ex.getMessage)
+            if (ex.isInternalError) {
+              val programContext = getProgramContext(environment.user, environment)
+              throw new CompilerServiceException(ex, programContext.dumpDebugInfo)
+            } else {
+              ExecutionRuntimeFailure(ex.getMessage)
+            }
           }
         } else {
           // Unexpected error. For now we throw the PolyglotException.
```
bgaidioz, line 415: When is that reached?

snapi-client/src/main/scala/raw/client/rql2/truffle/Rql2TruffleCompilerService.scala
```diff
@@ -404,7 +404,12 @@ class Rql2TruffleCompilerService(maybeClassLoader: Option[ClassLoader] = None)(
             }
             ExecutionValidationFailure(errors.to)
           } else {
-            ExecutionRuntimeFailure(ex.getMessage)
+            if (ex.isInternalError) {
```
bgaidioz, line 407: I noticed all the plain internal bug exceptions (java exceptions) have that flag true, and our RawException (which we get when a file doesn't parse), doesn't. XmlPackageTest is failing without this check because the code throws the RawException to the layer above.

snapi-truffle/src/main/java/raw/runtime/truffle/RawLanguage.java
```diff
@@ -167,21 +167,16 @@ protected CallTarget parse(ParsingRequest request) throws Exception {
...
-                    ex.getMessage(),
-                    JavaConverters.asScalaBufferConverter(
-                            java.util.Collections.singletonList(
-                                new ErrorRange(ex.position(), ex.position())))
-                        .asScala()
-                        .toList())));
-      } else {
-        throw e;
-      }
+    } catch (CompilerParserException ex) {
```
bgaidioz, line 170: Putting a breakpoint there on a parse error, I see the exception itself is a `CompilerParserException`.

snapi-truffle/src/main/java/raw/runtime/truffle/tryable_nullable/Tryable.java
```diff
@@ -24,7 +25,7 @@ public static boolean isSuccess(Object value) {
   }
 
   public static String getFailure(Object value) {
-    if (!isFailure(value)) throw new RuntimeException("not a failure");
+    if (!isFailure(value)) throw new RawTruffleRuntimeException("not a failure");
```
bgaidioz, line 28: That was a typo, this exception has to be a language exception.

snapi-truffle/src/main/java/raw/runtime/truffle/runtime/exceptions/RawTruffleRuntimeException.java
```diff
@@ -46,4 +51,9 @@ public RawTruffleRuntimeException(Exception ex, Node location) {
     super(ex.toString(), location);
     //        this.location = location;
   }
+
+  @ExportMessage
+  public ExceptionType getExceptionType() {
+    return ExceptionType.EXIT;
```
bgaidioz, line 57: RUNTIME?

miguelbranco80 (approved):

PR #294, opened 2023-12-07
Remove dependencies and instead patch JARs dynamically

miguelbranco80:
This is my first contribution towards RD-10078: Find a more sustainable way to patch Automated Module Names in our Snapi dependencies

This removes jackson-module-scala, scala-logging.
Instead it patches those two in place plus jwt-core, which also required patching.


deps/others/manifest.txt
```diff
@@ -0,0 +1 @@
+Automatic-Module-Name: jwt.core
```
bgaidioz, line 1: Maybe that file should be in the `.gitignore`? I'm asking because I noticed its content is changing during compilation. It seems to be the latest package compiled or something random like that.

datYori:
Also we add the fact that if anything changes in this folder we make sure to sync our new hosting solution for this kind of deps in order to make sure the main repo can get them at build time

bgaidioz (approved):

PR #296, opened 2023-12-11
Fixed RD-10235: SyntaxAnalyzer failing after the flexible parser succeeds

bgaidioz:


miguelbranco80 (approved):
This PR should probably go in before LSP/parser PR so we get those tests.

PR #298, opened 2023-12-11
RD-10252: Fix desugaring when ValueParam are unexpectedly nullable/tryable

bgaidioz:
Added tests as `TestPackage`, that hit the bug. There are four failures without the patch.


torcato (approved):
Ok after chatting with you, the bug is with
```
outProps ++= report.extraProps
```
And the tests are triggering this code path.

PR #305, opened 2023-12-18
RD-10265: UTF-8 character escape failing

bgaidioz:
Note that this patch is _adding_ a dependency to `org.apache.commons.text` that has a utility performing the string escapes we need. Our `RawUtils.escape`/`descape` do not support dealing with UTF-8 escapes.


torcato (approved):
LGTM

PR #310, opened 2023-12-26
RD-10350 Add optional schema to PostgreSQL credential.

miguelbranco80:


bgaidioz:
We merge it after the SQL changes.

PR #316, opened 2024-01-11
jvm compilation failure flag for tests

datYori:


bgaidioz (approved):
I see that's only for tests.

Alexzerntev (approved):

PR #320, opened 2024-01-12
RD-10459: Fix the SQL connection pool

bgaidioz:
I misunderstood the usage of JDBC connection pools. This is a better implementation. The bug couldn't be reproduced with that implementation.

Note we add a new dependency.


torcato (approved):
LGTM

PR #323, opened 2024-01-15
RD-3852 SourceException doesn't interrupt auto-inference

torcato:
Removed the function  `withErrorHandling` in the inferrer and added the error handing in each inferrer.
The diff looks big but its no so much changing in each file.


bgaidioz (approved):

torcato:
\publish

PR #327, opened 2024-01-16
RD-10449: Add a new type of http credentials for Salesforce

stthanos:
We are introducing a new form of HTTP credentials for Salesforce. Presently, these credentials can only be utilized through the SQL compiler.

### DIFF
--- a/snapi-frontend/src/main/scala/raw/creds/api/Credentials.scala
+++ b/snapi-frontend/src/main/scala/raw/creds/api/Credentials.scala
@@ -152,7 +152,7 @@ final case class Secret(name: String, value: String) extends Credential
 object HttpCredentialType extends Enumeration {
   type CredentialType = Value
 
-  val ClientCredentials, Token, UserPass = Value
+  val ClientCredentials, Token, UserPass, Salesforce = Value
 
   def apply(name: String): Value = {
     values
@@ -170,7 +170,8 @@ object HttpCredentialType extends Enumeration {
   Array(
     new JsonType(value = classOf[BasicAuthCredential], name = "basic-auth"),
     new JsonType(value = classOf[TokenCredential], name = "oauth2-token"),
-    new JsonType(value = classOf[ClientCredentialsCredential], name = "oauth2-client-credentials")
+    new JsonType(value = classOf[ClientCredentialsCredential], name = "oauth2-client-credentials"),
+    new JsonType(value = classOf[SalesforceCredential], name = "salesforce-credential")
   )
 )
 sealed trait NewHttpCredential extends Credential {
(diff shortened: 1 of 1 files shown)


snapi-frontend/src/main/scala/raw/creds/api/Credentials.scala
```diff
@@ -210,10 +222,18 @@ case class HttpCredentialId(name: String, credType: HttpCredentialType.Value)
...
-    new JsonType(value = classOf[CustomHeaderToken], name = "custom-header")
+    new JsonType(value = classOf[CustomHeaderToken], name = "custom-header"),
+    new JsonType(value = classOf[SalesforceAuth], name = "salesforce-auth")
   )
 )
 sealed trait NewHttpAuth
 final case class BearerToken(token: String, options: Map[String, String]) extends NewHttpAuth
 final case class BasicAuth(username: String, password: String, options: Map[String, String]) extends NewHttpAuth
 final case class CustomHeaderToken(header: String, token: String, options: Map[String, String]) extends NewHttpAuth
+final case class SalesforceAuth(
```
miguelbranco80, line 233: I'm not sure this is correct. Can't this be a "OAuth2Provider" instead, as opposed to a top-level credential? Isn't SalesForce fundamentally just a type of OAuth2 provider? (Refer to class OAuth2Provider and how it implements various types like Zoho, etc)

miguelbranco80 (commented):
An important question!

snapi-frontend/src/main/scala/raw/creds/api/Credentials.scala
ppolydoras: @miguelbranco80  Salesforce steampipe plugin does not follow the OAuth authentication flow. The security token which appears to be part of the salesforce plugin configuration is something completely static and dependent on the user, not the origin server (Salesforce), so it's not a client secret. Right now, Salesforce connector/ cannot be used by the Snapi compiler/runner as an oauth credential and it's only usable by the SQL compiler. So, it felt closer to a Basic HTTP Auth credential. 

Please note that if we want to implement Salesforce connector accessible by Snapi (aka Salesforce.Account(...)), we have 2 options:
- use FDW as a data source provider and basically use JDBC behind the scenes (more preferable)
- consume Salesforce API, using OAuth flow and addressing all the tricky integration details

miguelbranco80: Ok. Let's see if we can define a framework for these, to know where to put things going forwarded.

Here's the current state:
- In HTTP, there are 3 types of credentials: basic auth, oauth token, oauth client credentials. There are no other types fundamentally of HTTP authentication in existence.
- For oauth, we have different protocols per provider.

So what is Salesforce? If not OAuth, then I'd say it's just a new type of credential? In that case, it's not a variant of HTTP authentication, because all the Snapi packages that use HTTP - e.g. do GET, POST, etc - would have to support "Salesforce HTTP" and that is non-sensical I'd say.

So the alternative is to add a new top-level credential type. This is more work with the current interface. But it's easy copy/paste work that takes a few minutes :) So I'd say to add it like we add Dropbox or what not: a bunch of new top-level methods to handle "Salesforce credential", which is its own type of credential and not a variant of HTTP (which would mean it'd have to support all HTTP operations) nor a variant of OAuth (which would mean it'd have to support all OAuth operations).

I can schedule a quick chat with @stthanos to show how this PR should be changed.

PR #333, opened 2024-01-22
Fixing dependency issues by creating maven projects

Alexzerntev:
- Added maven project for truffle and parser to resolve dependency issues. To keep CI/CD sbt configurations scripts and plugin, the new projects are dependencies for the initial sbt project. So there are snapi-parser that depends on snapi-parser-mvn and snapi-truffle that depends on snapi-truffle-mvn. At some point in the future, we should remove the sbt projects and keep only the maven one


Alexzerntev:
@datYori could you please take a look on how we need to proceed with this?

datYori (changes requested):
CI needs adaptation. I guess I should do them

datYori:
done through https://github.com/raw-labs/snapi/pull/383

PR #334, opened 2024-01-22
Fixed RD-10499: Unsupported jsonb

bgaidioz:
Also fixed a couple of issues regarding null values and CSV rendering null fields wrong.


sql-client/src/main/scala/raw/client/sql/writers/TypedResultSetCsvWriter.scala
```diff
@@ -92,6 +99,16 @@ class TypedResultSetCsvWriter(os: OutputStream, lineSeparator: String) {
...
       case _: RawStringType => gen.writeString(v.getString(i))
+      case _: RawAnyType => v.getMetaData.getColumnTypeName(i) match {
+          case "jsonb" =>
+            val data = v.getString(i)
+            // RawAnyType cannot be nullable, but jsonb can be null.
+            if (v.wasNull()) gen.writeNull()
+            else {
+              gen.writeString(data)
+            }
+          case _ => gen.writeString(v.getString(i))
```
bgaidioz, line 110: not supported maybe?

torcato (approved):
LGTM

PR #337, opened 2024-01-23
RD-10503: CSV Truffle writers aren't writing null values correctly

bgaidioz:
I added tests that broke before the fixes.

That's because writers are writing items as if they were in an array. If switching to "object", tests pass.


torcato (approved):
Ok, so using write-object instead of the write array methods.
And some tests. Approved!

PR #338, opened 2024-01-24
RD-10439 data duplication on MySQL.InferAndRead

torcato:
The problem was not using the database when asking the metadata for tables.

I had also to change the RDBMS clients as we were using a lot of times the credential name instead of the database base name.
So added a val `database: Option[String]` to the base jdbc client and changed the way of inferring.


snapi-frontend/src/main/scala/raw/sources/jdbc/api/JdbcClient.scala
```diff
@@ -58,6 +58,8 @@ abstract class JdbcClient()(implicit settings: RawSettings) extends StrictLoggin
 
   def vendor: String
 
+  def database: Option[String]
```
bgaidioz, line 61: Could you please comment why it's optional. Thanks.

bgaidioz (approved):

PR #339, opened 2024-01-25
RD-10522 Truffle execution error in recursive function call

Alexzerntev:
- fixed the recursion bug and list take semantics


bgaidioz (commented):
Since you're fixing the Take issue. Can you add a test, in both `ListPackageTest` and `CollectionPackageTest` (maybe we don't follow the right semantic there too?)

bgaidioz (approved):

PR #344, opened 2024-01-26
RD-10500 stop creating unnecessary closures

Alexzerntev:
- Fixed collection Build bug.
- Separated Closures from NonClosures, from lambdas from RecFunctions


snapi-truffle/src/main/java/raw/runtime/truffle/ast/expressions/function/LambdaNode.java
```diff
@@ -0,0 +1,39 @@
...
+  @CompilerDirectives.CompilationFinal private Lambda lambda;
+
+  public LambdaNode(Function f) {
+    callTarget = f.getCallTarget();
+  }
+
+  @Override
+  @ExplodeLoop
+  public Object executeGeneric(VirtualFrame virtualFrame) {
+    if (lambda == null) {
```
bgaidioz, line 34: With comments explaining the specificity of that type of closure, you can explain why it's then possible to cache that `lambda`.

snapi-truffle/src/main/java/raw/compiler/snapi/truffle/compiler/SnapiTruffleEmitter.java
```diff
@@ -128,7 +127,8 @@ protected StatementNode emitMethod(Rql2Method m) {
         ExpressionNode[] defaultArgs = JavaConverters.asJavaCollection(fp.ps()).stream()
                 .map(p -> p.e().isDefined() ? recurseExp(p.e().get()) : null)
                 .toArray(ExpressionNode[]::new);
-        ClosureNode functionLiteralNode = new MethodNode(m.i().idn(), f, defaultArgs);
+        // The NonClosureNode materializes the frame if there are free variables.
```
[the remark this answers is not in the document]
bgaidioz: The method may be implemented as a `Lambda` if it has no free vars + no optional args. And repose is going to do the right thing (pass all parameters in the right order).

snapi-truffle/src/main/java/raw/runtime/truffle/ast/expressions/function/MethodNode.java
```diff
@@ -14,51 +14,64 @@
...
         }
+        function =
+            new Closure(
+                this.parentFunction,
+                defaultArguments,
+                hasFreeVars ? virtualFrame.materialize() : null);
+      }
+      // Otherwise it is a lambda
+      else {
+        return new Lambda(this.parentFunction.getCallTarget());
```
bgaidioz, line 68: Shouldn't it set `function` to that `new Lambda` and get it registered?

bgaidioz (approved):

PR #348, opened 2024-01-31
RD-10538: Nullable columns sometimes reported as non-nullable

bgaidioz:
The code didn't check if the status was 'unknown' (in which case, we should make it nullable).

A test was added that reports a column has unknown nullability.

An existing test had to be fixed since we were asserting its output had a certain nullability. Since we changed the code, the nullability of its column changed.

Also, an existing test was still assuming decimals were surrounded with double quotes (which was the case before RD-10527 got fixed).


torcato (approved):
LGTM

PR #350, opened 2024-02-01
RD-10542: Suggest adding credentials when a relation isn't found

bgaidioz:
Since schemas and tables appear only when adding credentials, this would invite users to add credentials when such error occurs.

### DIFF
--- a/sql-client/src/main/scala/raw/client/sql/ErrorHandling.scala
+++ b/sql-client/src/main/scala/raw/client/sql/ErrorHandling.scala
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2023 RAW Labs S.A.
+ *
+ * Use of this software is governed by the Business Source License
+ * included in the file licenses/BSL.txt.
+ *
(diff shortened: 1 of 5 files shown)


sql-client/src/main/scala/raw/client/sql/ErrorHandling.scala
```diff
@@ -0,0 +1,43 @@
...
+import raw.client.api.{ErrorMessage, ErrorPosition, ErrorRange}
+
+import java.sql.SQLException
+
+object ErrorHandling extends StrictLogging {
+
+  def asErrorMessage(source: String, exception: SQLException): List[ErrorMessage] = {
+    val message = exception.getSQLState match {
+      case "42P01" => // undefined table, add a hint
+        exception.getMessage + "\nDid you forget to add credentials?"
```
bgaidioz, line 25: @ppolydoras would like the message to be picked from config to have the possibility to change it.

torcato (approved):
LGTM

PR #352, opened 2024-02-04
RD-10562 Implement OSR for iterators

Alexzerntev:
- Updated aggregations to use OSR
- Updated lists to use OSR
- Updated joins to use OSR
- Updated iterables to use OSR
- Updated Json list reader to use OSR
- RecordObject now extends Dynamic Truffle object


snapi-truffle/src/main/java/raw/runtime/truffle/ast/expressions/iterable/list/osr/OSRToArrayNode.java
```diff
@@ -0,0 +1,77 @@
...
+  public boolean executeRepeating(VirtualFrame frame) {
+    // ignored
+    return false;
+  }
+
+  public boolean shouldContinue(Object returnValue) {
+    return returnValue == this.initialLoopStatus() || currentIdx < llist.size();
+  }
+
+  public Object executeRepeatingWithValue(VirtualFrame frame) {
```
bgaidioz, line 51: The doc of `executeRepeatingWithValue` says:  "return a value `v` satisfying `shouldContinue(v) == true` if the method should be executed again to complete the loop and any other value if it must not." (That last part I interpret as "any other value _that doesn' t satisfy `shouldContinue`_.)

I see `shouldContinue`'s code is comparing its parameter to `initialLoopStatus`, which is a status, an object like `CONTINUE_LOOP_STATUS`. But with `executeRepeatingWithValue` returning the result array, it will perform the same test with that array. I understand that will evaluate to `false` and only the check on `currentIdx` is going to matter. But that feels like not the intention, to spend the whole loop execution comparing the array to the "enum".

I interpret the `OptimizedOSRLoopNode` logic as if the `return` value of the node _is a status value_ (and can be custom if using the `withValue` version: `BREAK`, `TOOK_TOO_LONG`, `INTERRUPTED`, etc.) and the way a loop would return an actual result would be either as a side-effect (by updating a mutable object passed as a parameter) or by exposing a `getResult()` method that returns an object it would have allocated internally. (e.g. all these `OSR...Node` classes would compute the result internally, and inherit an abstract `class` (which itself inherits `RepeatingNode`), that would require a `getResult()` method. Which is called after `execute`?
```java
Node existNode = new ExistNode();
existNode.execute(); // returns the status, in our case `true/false`, no need for a custom status?
return existNode.getResult();
```

I looked in `TruffleScheme` code a bit everywhere to find how loop-nodes that return a value are implemented. Scheme has plenty, like `length`, `map`, etc. All the expected ones in fact work with regular `for` loops instead of using `RepeatingNode` (they probably didn't fix them yet..?). There's one place I found the `TailRecursiveLoop` (which _is_ a `RepeatingNode`) called, and it goes with [calling `execute` (ignoring its result) and picking the result from elsewhere](https://github.com/horakivo/TruffleScheme/blob/76ea42fadb4de743163190873cf3281ad6dd16de/TruffleScheme/language/src/main/java/com/ihorak/truffle/node/callable/TCO/TailRecursiveRootNode.java#L39). In that specific case, the node returning a value through `execute` (`TailRecursiveRootNode`) is initialized with an index where to write the result too, and the looping node (`TailRecursiveLoopNode`) is getting it as a parameter. Internally, it writes in that index, while looping with a regular status implementation. Which is quite like it would be initialized with a mutable object and update it as the loop runs.

bgaidioz (commented):
A question about the design. It felt to me like the implementation doesn't use the `RepeatingNode` correctly. But I only found a specific example to compare this to.

snapi-truffle/src/main/java/raw/runtime/truffle/ast/osr/filter/OSRListFilterConditionNode.java
```diff
@@ -0,0 +1,46 @@
...
+  private final int generatorSlot;
+
+  public OSRListFilterConditionNode(int generatorSlot) {
+    this.generatorSlot = generatorSlot;
+  }
+
+  @Override
+  public Object executeGeneric(VirtualFrame frame) {
+    Object generator = frame.getAuxiliarySlot(generatorSlot);
+    if (!hasNextNode.execute(this, generator)) {
```
Alexzerntev, line 36: todo, remove the if

bgaidioz (approved):

Alexzerntev:
To be removed, will be merged through https://github.com/raw-labs/snapi/pull/379

PR #354, opened 2024-02-05
RD-10004 improve-error-propagation-for-lsp-in-the-compiler-interfaces

Alexzerntev:
Updated Errors class, added:
- code as a type of error message
- tags (deprecated etc..)
- severity 

These information is corespondent to what monaco editor expects

### DIFF
--- a/client/src/main/scala/raw/client/api/Errors.scala
+++ b/client/src/main/scala/raw/client/api/Errors.scala
@@ -18,8 +18,17 @@ import raw.utils.RawException
  * Used for errors that are found during semantic analysis.
  * @param message The error message.
  * @param positions The positions where the error occurred.
+ * @param severity The severity of the error. 1 = Hint, 2 = Info, 4 = Warning, 8 = Error
+ * @param code An optional error code.
+ * @param tags Indication for the error Unnecessary = 1, Deprecated = 2
  */
-final case class ErrorMessage(message: String, positions: List[ErrorRange])
+final case class ErrorMessage(
+    message: String,
+    positions: List[ErrorRange],
+    severity: Int = 8,
+    code: Option[String] = None,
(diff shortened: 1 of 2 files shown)


client/src/main/scala/raw/client/api/Errors.scala
```diff
@@ -18,8 +18,17 @@ import raw.utils.RawException
...
  * @param positions The positions where the error occurred.
+ * @param severity The severity of the error. 1 = Hint, 2 = Info, 4 = Warning, 8 = Error
+ * @param code An optional error code.
+ * @param tags Indication for the error Unnecessary = 1, Deprecated = 2
  */
-final case class ErrorMessage(message: String, positions: List[ErrorRange])
+final case class ErrorMessage(
+    message: String,
+    positions: List[ErrorRange],
+    severity: Int = 8,
```
miguelbranco80, line 28: Suggest you strict the API. By making it a number, I guarantee one of us will add an 'id' to it by accident ;) Make it an scala enum and use that as the API. Can use the underlying numbers in the enum definition.

client/src/main/scala/raw/client/api/Errors.scala
```diff
@@ -18,8 +18,17 @@ import raw.utils.RawException
...
+    code: Option[String] = None,
+    tags: List[Integer] = List.empty
```
miguelbranco80, line 30: BTW, this is Integer, not Int. In any case, same comment as before: use a scala enum.

bgaidioz (approved):

Alexzerntev:
To be closed, the final implementation will be provided here: https://github.com/raw-labs/snapi/pull/359

bgaidioz (approved):

PR #355, opened 2024-02-05
RD-10465 sql when no credentials

torcato:


sql-client/src/main/scala/raw/client/sql/SqlConnectionPool.scala
```diff
@@ -11,17 +11,45 @@
...
-import raw.utils.{AuthenticatedUser, RawSettings}
+import raw.utils.{AuthenticatedUser, RawConcurrentHashMap, RawException, RawSettings}
 
 import java.sql.SQLException
 import java.util.concurrent.TimeUnit
 import scala.collection.mutable
 
-class SqlConnectionPool(settings: RawSettings) {
+//class SqlClientMissingCredentialsException(cause: Throwable) extends RawException("add credentials to use SQL", cause)
+class SqlClientMissingCredentialsException(cause: Throwable)
```
bgaidioz, line 24: I think this `Exception` can be deleted.

sql-client/src/main/scala/raw/client/sql/SqlConnectionPool.scala
```diff
@@ -11,17 +11,45 @@
...
+    extends RawException("to properly execute SQL queries, first add credentials", cause)
+
+class SqlConnectionPool(settings: RawSettings) extends StrictLogging {
 
-  // one pool of connections per DB (which means per user).
   private val pools = mutable.Map.empty[String, HikariDataSource]
+
+  // Default pool for users that do not have a DB yet
+  val defaultPool = {
```
bgaidioz, line 32: I suppose it can be `private`.

sql-client/src/main/scala/raw/client/sql/SqlConnectionPool.scala
```diff
@@ -44,9 +81,26 @@ class SqlConnectionPool(settings: RawSettings) {
...
         config.setUsername(readOnlyUser)
         config.setPassword(password)
-        val pool = new HikariDataSource(config)
-        pools.put(db, pool)
-        pool
+        try {
+          val pool = new HikariDataSource(config)
+          // remove the key from the map, we will now connect only to the user DB
+          logger.debug(s"Connected to user deleting default entry for $db")
+          defaultPoolsLastTime.remove(db)
```
bgaidioz, line 88: Is it possible that it crashes if `db` isn't found? (That would be the case if creating credentials before going to SQL, or if restarting the system the day after a user has gone through the usual steps.)

torcato: If the value is not in the map then the remove does nothing.
Looked at the documentation of ConcurrentHashMap, the thing used by that class:
```
Removes the key (and its corresponding value) from this map. This method does nothing if the key is not in the map.
```

bgaidioz (approved):

PR #359, opened 2024-02-08
RD-10005 flag a warning when environment secret not set

Alexzerntev:
- Added support for compilation messages


snapi-frontend/src/main/scala/raw/compiler/base/errors/Errors.scala
```diff
@@ -13,26 +13,69 @@
...
+      case e: ErrorCompilationMessage => ErrorMessage(ErrorsPrettyPrinter.format(e), range, e.code)
+      case w: WarningCompilationMessage => WarningMessage(ErrorsPrettyPrinter.format(w), range, w.code)
+      case i: InfoCompilationMessage => InfoMessage(ErrorsPrettyPrinter.format(i), range, i.code)
+      case h: HintCompilationMessage => HintMessage(ErrorsPrettyPrinter.format(h), range, h.code)
+      case _ => throw new AssertionError("Unknown message type")
+    }
+  }
+}
+
+trait CompilationMessage extends BaseNode {
```
miguelbranco80, line 31: Change to sealed trait.
Then, you do not need case _ in line 26.
Suggest renaming to CompilerMessage, since it's not stricly a compilation message, but rather a message of the compiler. For instance, it's also used during validation, and that doesn't involve compilation.

snapi-frontend/src/main/scala/raw/compiler/base/errors/Errors.scala
```diff
@@ -13,26 +13,69 @@
 package raw.compiler.base.errors
 
 import raw.compiler.base.source._
-import raw.client.api.ErrorMessage
+import raw.client.api.{ErrorMessage, ErrorRange, HintMessage, InfoMessage, Message, WarningMessage}
+import raw.compiler.rql2.errors.ErrorsPrettyPrinter
 
-trait BaseError extends BaseNode {
+object CompilationMessageMapper {
+  def toMessage(err: CompilationMessage, range: List[ErrorRange]): Message = {
```
miguelbranco80, line 20: This helpers bridges two interfaces. Where is it consumed? If it's consume only by the compiler service interface, then it should move then. raw.compiler.base ideally only uses compiler internals, and the least bit possible of public interfaces.
But this helper, which produces CompilerService API out of the Compiler API, feels like it should be more closer to the CompilerService classes/packages, and not inside raw.compiler.base ...

Alexzerntev: It is also used used in here : https://github.com/raw-labs/snapi/pull/359/files#diff-3e266db89cee304ea44c7cf8356e6fa6f106f11960539cf5a5db3b9eafe5a663

snapi-frontend/src/main/scala/raw/compiler/base/errors/Errors.scala
Alexzerntev: SourceError extends it in scala2Compiler

bgaidioz (approved):

PR #362, opened 2024-02-12
refactor(deps)(others)(build): improve jar path retrieval

datYori:
- Stop try to locate only coursier cache, just use the `update.value` infos
- allow to re-publish locally kiama


bgaidioz (approved):
I could compile on Mac.

PR #366, opened 2024-02-23
RD-10539 SELECT :v ::interval AS column fails with internal error

torcato:
In this PR:
* Added a way of checking if jdbc.OTHER is an interval or an json
* Added a validation for input parameters.
* Added an object SqlIntervals with functions to parse pgsql intervals and convert intervals to string.

### DIFF
--- a/sql-client/src/main/scala/raw/client/sql/NamedParametersPreparedStatement.scala
+++ b/sql-client/src/main/scala/raw/client/sql/NamedParametersPreparedStatement.scala
@@ -101,6 +101,17 @@ class NamedParametersPreparedStatement(conn: Connection, code: String)(implicit 
     plainCodeBuffer.toString()
   }
 
+  private def validateParameterType(tipe: RawType): Either[String, RawType] = tipe match {
+    case _: RawNumberType => Right(tipe)
+    case _: RawStringType => Right(tipe)
+    case _: RawBoolType => Right(tipe)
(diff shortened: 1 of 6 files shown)


sql-client/src/main/scala/raw/client/sql/NamedParametersPreparedStatement.scala
```diff
@@ -101,6 +101,17 @@ class NamedParametersPreparedStatement(conn: Connection, code: String)(implicit
...
 
+  private def validateParameterType(tipe: RawType): Either[String, RawType] = tipe match {
+    case _: RawNumberType => Right(tipe)
+    case _: RawStringType => Right(tipe)
+    case _: RawBoolType => Right(tipe)
+    case _: RawDateType => Right(tipe)
+    case _: RawTimeType => Right(tipe)
+    case _: RawTimestampType => Right(tipe)
+    case _: RawBinaryType => Right(tipe)
+    case _ => Left(s"Unsupported parameter type: $tipe")
```
bgaidioz, line 112: How does this render? I fear it renders with the name of the case class, something one wouldn't want to show to users.

torcato: Yes, it will.
But we do not really have a pretty printer.
Let me see what I can come up with

bgaidioz (approved):

PR #367, opened 2024-02-27
RD-10612 hovering on a parameter name doesn't return the parameter type + fails internally

torcato:
This was because the list of tokens was empty in the `hover` method of the `SqlCompilerService`.
In the end this was because the function `getIdentifierUnder` should have used a `>=` instead of an `=` to get the boundary of the token also

### DIFF
--- a/sql-client/src/main/scala/raw/client/sql/SqlCodeUtils.scala
+++ b/sql-client/src/main/scala/raw/client/sql/SqlCodeUtils.scala
@@ -200,7 +200,7 @@ class SqlCodeUtils(code: String) {
     val tokens = SqlCodeUtils.tokens(code)
     // Finds the corresponding token
     val maybeToken = tokens.find {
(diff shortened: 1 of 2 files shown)


sql-client/src/main/scala/raw/client/sql/SqlCodeUtils.scala
```diff
@@ -200,7 +200,7 @@ class SqlCodeUtils(code: String) {
     val tokens = SqlCodeUtils.tokens(code)
     // Finds the corresponding token
     val maybeToken = tokens.find {
-      case (token, pos) => pos.column <= p.column && (pos.column + token.length) > p.column && pos.line == p.line
+      case (token, pos) => pos.column <= p.column && (pos.column + token.length) >= p.column && pos.line == p.line
```
bgaidioz, line 203: Yes, it's now the same as `getIdentifierUpTo`.

bgaidioz (approved):

bgaidioz (approved):

bgaidioz (approved):

PR #369, opened 2024-03-01
Support multiple Truffle Engines incl. independent settings.

miguelbranco80:


snapi-truffle/src/main/java/raw/runtime/truffle/RawLanguageCache.java
```diff
@@ -78,19 +83,19 @@ private Value get(AuthenticatedUser user) {
         });
   }
 
-  public SourceContext getSourceContext(AuthenticatedUser user) {
-    return get(user).getSourceContext();
+  public SourceContext getSourceContext(AuthenticatedUser user, RawSettings rawSettings) {
+    return get(user, rawSettings).getSourceContext();
```
bgaidioz, line 87: In a production setup, there's only a single `rawSettings` and each user has one source context, inferrer, etc.. After that change, fetching these user's source context, inferrer, needs computing the hash of the `AuthenticatedUser` (it was the case before) and of that single `rawSettings` object (the search is in a hash map) each time, although it's technically the same and only one.

A `RawSettings` has quite a lot of keys. Last time we would parse the settings once per query and it would be visible for short queries. I suppose the computation of its hash is quicker, but since it's done in these methods, is there any condition/query under which computing this hash would be done a lot (and perhaps impact performance)?

Just trying to come up with cases I only think of: a query that infers a lot of inlined strings or files, or one that retrieves `Environment.Secret` items at runtime in a loop.

client/src/main/scala/raw/client/api/CompilerService.scala
```diff
@@ -34,59 +34,48 @@ final class CompilerServiceException(
...
-          //  "-Dgraal.DumpPath=/tmp/graal_dumps",
-          //  "-Dgraal.PrintGraph=Network",
-          // https://www.graalvm.org/latest/graalvm-as-a-platform/language-implementation-framework/Options/
-        }
-        engine = Engine.newBuilder().allowExperimentalOptions(true).options(options).build()
+  private val enginesLock = new Object
+  private val enginesCache = mutable.HashMap[RawSettings, Engine]()
+
+  // Return engine and a flag indicating if the engine was created.
+  def getEngine()(implicit settings: RawSettings): (Engine, Boolean) = {
```
bgaidioz, line 41: Similar question. How is the hash of a `RawSettings` computed?

snapi-truffle/src/main/java/raw/runtime/truffle/RawLanguageCache.java
miguelbranco80: RawSettings is simply a class, and because we are not overridding equals and not overriding hashCode, this will be the default fast computation.
Now, it could mean that for two RawSettings with "same data" - but different objects - they will not hash together in the same bucket. But here we want to go for speed and not optimize for a case that won't happen in prod.

bgaidioz (approved):

PR #372, opened 2024-03-05
RD-10698 runner crash report scala match error

Alexzerntev:
Fixed


snapi-client/src/test/scala/raw/compiler/rql2/tests/lsp/LspDefinitionTest.scala
```diff
@@ -175,4 +175,13 @@ trait LspDefinitionTest extends CompilerTestContext {
     assert(goToDefinitionResponse.position.isEmpty)
   }
 
+  test("RD-10698 crush fix") { _ =>
```
bgaidioz, line 178: Can you write "crash"? Crush is something else.

bgaidioz (approved):

PR #374, opened 2024-03-07
RD-10741 Improve error handling in the REST Client.

miguelbranco80:
* Adds new exceptions for bad responses so that messages are clearly defined
* Centralizes automatic retries on 503 in a single method
* Remove accidentally passing bodies as exception messages for auth error

Note that this removes the "fundamentally broken" handleUnexpected method, which is still used by the Runner, so a small fix will be necessary there.


miguelbranco80:
This needs an addition which is to remove raw.auth altogether - Exceptions and Classes -, since it's not longer necessary!

torcato (approved):
LGTM

PR #382, opened 2024-03-14
Integration of the SQL parser (fixes: RD-10712, RD-10712, RD-10681, RD-10612)

bgaidioz:
Fixes:
* RD-10712: Hover (and autocomplete and all) mistakenly search for a parameter name in the schema/tables metadata,
* RD-10505: Comments _after_ a statement that ends with `;` are parsed as a second statement (not allowed),
* RD-10681: JSON literal strings badly parsed,
* RD-10612: Hover crash.


torcato (approved):
We did a review, tested together

PR #383, opened 2024-03-14
refactor(snapi-parser)(build): build with maven

datYori:
We move from `sbt` to `maven` in order to build `snapi-parser` component

Since `snapi-parser` is not depending on any other `snapi` component we take advantage of this task to split the publish job.

We kept the same way of following the git tags for versioning

We now use the `antlr4-maven-plugin` directly to generate our parsers

We remove every sbt related stuff

RD-10756 internal ref


Alexzerntev:
Could you please rename it from "snapi-parser" to "parsers" or "raw-parsers" (or maybe you can come up with a better name) because now it also contain sql-parser except for snapi-parser.

PR #384, opened 2024-03-15
RD-10677 Support `getUserDB` capability in `CredentialsService` API

ppolydoras:


snapi-frontend/src/main/scala/raw/creds/client/ClientCredentialsService.scala
```diff
@@ -12,16 +12,32 @@
...
 import raw.rest.client.APIException
+import raw.utils.{AuthenticatedUser, RawSettings}
 
 import java.net.URI
+import java.util.concurrent.TimeUnit
+
+object ClientCredentialsService {
+  private val SERVER_ADDRESS = "raw.creds.client.server-address"
+  private val CACHE_FDW_SIZE = "raw.creds.client.fdw-db-cache.size"
+  private val DEFAULT_CACHE_FDW_SIZE = 1000
```
bgaidioz, line 33: In our codebase, default setting values are often defined as values in `reference.conf` files. Which plays the role of the `.getOrElse` that is done below. You prefer it like that?

bgaidioz (approved):

PR #386, opened 2024-03-15
RD-10537: Optional parameters in SQL

bgaidioz:
If accepting default values, we need to check values are compiling, but also that their type is the same as what `@type` specifies (if it is there).

The way it's implemented is:

* `@type` declarations are walked, validated (for what we support), set as nullable. This is collected in `userSpecifiedTypes`.
* `@default` declarations are walked the same way. One by one their expressions are typed and lead to a failure if an existing `@type`
   contradicts that type. If not, they are executed and their value is turned into a `RawValue`. This happens in `userSpecifiedDefaultValues`.

Both are merged into `declaredTypeInfo`, which is a `Map[String,
Either[ErrorMessage, QueryParamInfo]]`, where `QueryParamInfo` is a tuple
(`PostgresType`, `Option[RawValue]`). That permits to have a mapping of parameter names to types that are either:
* known only from `@type`, possibly `Left` if not supported
* known only from `@default` (we inferred the type from the expression, possibly `Left` if not supported or typo in the code)
* known from `@type` and `@default` (possibly `Left` if they aren't the same)

That's pretty much equivalent to rely on the user entered `@type` as it happens
today. Just that now we've confronted `@type` and `@default`, and also we have
the default values computed. And we have already computed the `ErrorMessage` with the
proper underlining in case. (Before it would be a `String`.)

The final query parameter metadata that relies on Postgres' inference and typechecking is implemented as before. We still replace parameters by:

* `?` when `@type` or `@default` weren't specified
* `?::<tipe>` when we've validated the types found with `@type` and/or `@default`
* `?` when `@type`/`@default` lead to a `Left`.

Since we collect all the type info with already flagged errors when needed, that simplifies a bit the final query metadata computation (see `collectParamInfo`) which now relies on that processed type info instead of jumping to `@type`. That
last part still reports the inferred parameter types either using the user entered info (merge of `@type` and `@default`) or the inferred types.

I also changed the interface to run the statement. Instead of exposing step-by-step parameter setters + execute, now it takes parameters as `RawValue` and returns the `ResultSet`. Internally it picks the right setter, which is also used to set default values.

A _very important_ point to check is a parser change I had to do: by checking the positions' offsets, I noticed in certain cases Kiama would return `None` for a position offset. The reason is the parser would store a wrong column
offset. (This will be hit if disabling the patch.) What happens is:
* `EOF` is accepted in several parts of the parser, not only after `prog`. One of the tests _ends_ with a single line comment, and `EOF` becomes part of it.
* When computing the position, the end position is obtained by picking the last token's position + adding its string length. When `EOF` is part of the comment, therefore of the statement, that logic leads to compute the string length of `EOF`
and that happens to _be_ `"<EOF>"`. Which leads the end position to be five characters too far (68 instead of 63).

And we can't compute the end position of the first valid statement, therefore we can't cut the SQL code stopping at the first statement.


torcato (approved):
We went over the code together. So approved.

PR #393, opened 2024-03-29
Fixed RD-10801: do not parse/read Json data that has duplicated record fields

bgaidioz:
Here's a fix to the fact our Json parser/reader can't load data that has duplicated fields (since it reads fields out of order, it wouldn't know which one to pick). Records with duplicated fields are considered as not supported, like functions, package, etc.

There was some refactoring of the `validateJsonType` function so that it can apply to both user entered types (which we can underline in the error) or the inferred entered type.

With that patch, `UnsupportedType` supports _hint_ which permits to flag the duplicate field in the error message returned to the user.

I also edited XML which cannot parse records with duplicate fields (duplicate tags lead to a collection).

As part of the fix, I noticed `Json.Print` was failing if its parameter wouldn't be supported by our reader. Which turns out to be slightly wrong: it should fail if its parameter isn't supported by our _writer_. Because it never _parses_ a string, it only _renders_ data as string. In particular, it deals with duplicate fields like our regular query writer does (they're renamed). I changed that too.

### DIFF
--- a/snapi-client/src/test/scala/raw/compiler/rql2/tests/truffle/regressions/TruffleRegressionTest.scala
+++ b/snapi-client/src/test/scala/raw/compiler/rql2/tests/truffle/regressions/TruffleRegressionTest.scala
@@ -73,3 +73,4 @@ import raw.testing.tags.TruffleTests
 @TruffleTests class RD10194TruffleTest extends TruffleCompilerTestContext with RD10194Test
 @TruffleTests class RD10220TruffleTest extends TruffleCompilerTestContext with RD10220Test
 @TruffleTests class RD10723TruffleTest extends TruffleCompilerTestContext with RD10723Test
+@TruffleTests class RD10801TruffleTest extends TruffleCompilerTestContext with RD10801Test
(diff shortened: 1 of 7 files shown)


snapi-client/src/test/scala/raw/compiler/rql2/tests/truffle/regressions/TruffleRegressionTest.scala
```diff
@@ -73,3 +73,4 @@ import raw.testing.tags.TruffleTests
 @TruffleTests class RD10194TruffleTest extends TruffleCompilerTestContext with RD10194Test
 @TruffleTests class RD10220TruffleTest extends TruffleCompilerTestContext with RD10220Test
 @TruffleTests class RD10723TruffleTest extends TruffleCompilerTestContext with RD10723Test
+@TruffleTests class RD10801TruffleTest extends TruffleCompilerTestContext with RD10801Test
```
bgaidioz, line 76: Question while reviewing: how is it working in CSV and RDBMS? CSV probably supports it (with infer and without) while RDBMS doesn't, but try to pass a duplicate field record to Postgres.Read.

bgaidioz: Try InferAndQuery with a join duplicating fields (a table joined to itself),

bgaidioz: CSV is working fine. For RDBMS I opened https://raw-labs.atlassian.net/browse/RD-10855

torcato (approved):
LGTM

PR #400, opened 2024-04-12
RD-10767 typealias in top level declarations

torcato:
The method  `getProgramDescription`  in  the `Rql2CompilerService` class is looping through all top level declarations.
After some discussion we decided to make the program-description return Option-Types for declarations (both in output types and parameters)


torcato:
\publish

bgaidioz (approved):

PR #404, opened 2024-04-18
refactor(build): one sbt def to rule them all

datYori:
We group all components in a single sbt build definition.

This is a beta ugly version with some code duplication.

Incoming:
- dynamic license header handling
- annotation processor as a generic plugin
- grammar and lexer generation as plugin
- better antl4 and truffle-processor jar management

internal ref RD-10838


Dimitris-Mil (approved):
ci lgtm

bgaidioz (approved):

PR #421, opened 2024-05-01
fix(sql-client): sources dep and require

datYori:
I think this was missing from https://github.com/raw-labs/snapi/pull/414

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -355,7 +355,8 @@ lazy val sqlClient = (project in file("sql-client"))
   .dependsOn(
     client % "compile->compile;test->test",
     snapiFrontend % "compile->compile;test->test",
-    sqlParser % "compile->compile;test->test"
+    sqlParser % "compile->compile;test->test",
+    sources % "compile->compile;test->test",
   )
   .settings(
     commonSettings,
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -355,7 +355,8 @@ lazy val sqlClient = (project in file("sql-client"))
   .dependsOn(
     client % "compile->compile;test->test",
     snapiFrontend % "compile->compile;test->test",
```
miguelbranco80, line 357: Good catch but actually we also need to remove the snapiFrontend dependency, since SQL should no longer be depending on Snapi.
If something still breaks on that change, then pass the PR to myself or to @bgaidioz

sql-client/src/main/java/module-info.java
```diff
@@ -23,6 +23,7 @@
   requires java.sql;
   requires com.zaxxer.hikari;
   requires raw.snapi.frontend;
```
miguelbranco80, line 25: Once we remove the snapiFrontend dependency, this line should also go.

build.sbt
datYori: we still need it as a test dependency because of `CredentialsTestContext` being part of `snapi-frontend`

miguelbranco80 (approved):

PR #423, opened 2024-05-07
refactor(snapi-parser)(sql-parser)(build): better build def

datYori:
- antlr4 custom tasks as plugin
- removes antlr4 jar from git repo
- download antlr4-complete jar as part of the plugin

internal ref: RD-10872


torcato (approved):
LGTM

PR #424, opened 2024-05-07
RD-10895: Salesforce credential expired leads to an unexpected error 500

bgaidioz:
It would be nice to test the change using an expired credential. What I believe happens is: the "connection error" that occurs due to the expired credential, occurs _at runtime_ for us, when running the SQL query. In the actual code logic, any `PSQLException` is assumed to be an internal failure, since the user query has been validated successfully. However a runtime error can be due to connection failures to the datasource (the case in the ticket), or "second level validation errors":
* the original source table is gone or was renamed: if so, our Postgres engine validates a user query reading a table that eventually doesn't exist,
* the table exists, but isn't readable,
* the password was changed, etc.

In this changeset, all PSQL errors (which include errors due to the remote source failing), are reported to the user.

### DIFF
--- a/client/src/main/scala/raw/client/api/CompilerService.scala
+++ b/client/src/main/scala/raw/client/api/CompilerService.scala
@@ -30,8 +30,9 @@ final class CompilerServiceException(
 
   def this(t: Throwable, debugInfo: List[(String, String)]) = this(t.getMessage, debugInfo, t)
 
-  def this(t: Throwable, environment: ProgramEnvironment) =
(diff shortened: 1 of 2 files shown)


torcato (approved):
Can we create an issue where we are catching the correct  postgres exceptions in the NamedParametersStatement class at liine 406?

bgaidioz:
> Can we create an issue where we are catching the correct postgres exceptions in the NamedParametersStatement class at liine 406?

Added a comment in our tech debt doc in Confluence.

PR #430, opened 2024-05-13
Fixed RD-10923: SQL time/timestamps are losing milliseconds

bgaidioz:


torcato (approved):
LGTM

sql-client/src/main/scala/raw/client/sql/writers/TypedResultSetJsonWriter.scala
```diff
@@ -102,11 +103,11 @@ class TypedResultSetJsonWriter(os: OutputStream) {
         val date = v.getDate(i).toLocalDate
         gen.writeString(dateFormatter.format(date))
       case _: RawTimeType =>
-        val time = v.getTime(i).toLocalTime
+        val time = v.getObject(i, classOf[LocalTime])
         val formatted = timeFormatter.format(time)
         gen.writeString(formatted)
       case _: RawTimestampType =>
-        val dateTime = v.getTimestamp(i).toLocalDateTime
+        val dateTime = v.getObject(i, classOf[LocalDateTime])
```
bgaidioz, line 110: This fails on expressions like `NOW()` because it's a `TIMESTAMPTZ` and the call to `.getObject` fails.

sql-compiler/src/main/scala/com/rawlabs/sql/compiler/writers/TypedResultSetCsvWriter.scala
```diff
@@ -143,7 +144,8 @@ class TypedResultSetCsvWriter(os: OutputStream, lineSeparator: String, maxRows:
         val date = v.getDate(i).toLocalDate
         gen.writeString(dateFormatter.format(date))
       case _: RawTimeType =>
-        val time = v.getTime(i).toLocalTime
+        val sqlTime = v.getTime(i)
```
bgaidioz, line 147: Should have the same patch as in JSON

bgaidioz: Done. Also took care of not displaying `.000` when there are zero milliseconds, since it's our convention in CSV.

PR #431, opened 2024-05-13
Fixed RD-10924: SQL @default parser in multiline does not accept more than one word

bgaidioz:
Added a test and verified it fails without the patch.


torcato (approved):
LGTM

PR #433, opened 2024-05-14
RD-10863 Improve Snapi docs on Http regarding 'args'

torcato:
Added the type for args and headers and added an example.
This is how the new docs look like:

image

### DIFF
--- a/snapi-frontend/src/main/scala/raw/compiler/rql2/builtin/HttpPackage.scala
+++ b/snapi-frontend/src/main/scala/raw/compiler/rql2/builtin/HttpPackage.scala
@@ -103,7 +103,8 @@ abstract class HttpCallEntry(method: String) extends EntryExtension {
       ExampleDoc(
         s"""Http.${method.capitalize}(
           |  "https://www.somewhere.com/something",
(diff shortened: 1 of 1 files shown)


snapi-frontend/src/main/scala/raw/compiler/rql2/builtin/HttpPackage.scala
```diff
@@ -179,13 +180,17 @@ abstract class HttpCallEntry(method: String) extends EntryExtension {
       ParamDoc(
         "args",
         TypeDoc(List("list")),
-        "The query parameters arguments for the HTTP request. Parameters are URL-encoded automatically. Any key/value pair with a null key or value will be omitted and won't be included in the URL parameters.",
+        """The query parameters arguments for the HTTP request.
+          |<br/>Parameters are a list of tuples (string, string) e.g. `[{"name", "john"}, {"age", "22"}]`. They are URL-encoded automatically.
```
bgaidioz, line 184: This string is also sent to the UI when it retrieves the doc, and shows as a pop-up right?

HTML is rendered straight like that? (Imagine that string gets escaped or something.)

bgaidioz (approved):

PR #439, opened 2024-06-05
Fixed RD-10971

bgaidioz:
The crash was hit because we use `token.toString` when the `token` can be `null`. Instead I used `String.valueOf(token)`.

Also added tests for bugs hit while debugging that.

### DIFF
--- a/snapi-compiler/src/test/scala/com/rawlabs/snapi/compiler/tests/regressions/RD10971Test.scala
+++ b/snapi-compiler/src/test/scala/com/rawlabs/snapi/compiler/tests/regressions/RD10971Test.scala
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2024 RAW Labs S.A.
+ *
+ * Use of this software is governed by the Business Source License
+ * included in the file licenses/BSL.txt.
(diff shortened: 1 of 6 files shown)


snapi-compiler/src/test/scala/com/rawlabs/snapi/compiler/tests/regressions/RD10971Test.scala
```diff
@@ -0,0 +1,176 @@
...
+    // TODO: RD-10971: the record is indeed parsed as a failed record, but we skip the following one (Jane)
+    it should evaluateTo(
+      """[{name: "Benjamin", birthYear: 1978},
+        |Error.Build("'birthYear': not found"),
+        |{name: "Jane", birthYear: 200}]""".stripMargin
+    )
+  }
+
+  // Fails because of RD-10971
+  ignore(
```
bgaidioz, line 57: Add a test that mimics the actual RD-1971 bug (parsing an empty string as a collection of records).

And document the fix that was tried.

torcato (approved):
LGTM

PR #450, opened 2024-06-18
Remove remaining classloaders and Eval from CompilerService interface

miguelbranco80:


python-client/src/main/scala/raw/client/python/PythonCompilerService.scala
```diff
@@ -333,19 +278,19 @@ class PythonCompilerService(engineDefinition: (Engine, Boolean), maybeClassLoade
...
-    ctx.initialize("python")
-    ctx.enter()
-    try {
-      f(ctx)
-    } finally {
-      ctx.leave()
-      ctx.close()
-    }
-  }
+//  private def withTruffleContext[T](
```
bgaidioz, line 281: I suppose you meant to delete that?

bgaidioz (approved):

PR #451, opened 2024-06-21
RD-11042 Refresh FDW databases

torcato:


sql-client/src/main/scala/raw/client/sql/SqlConnectionPool.scala
```diff
@@ -37,58 +33,52 @@ class SqlConnectionPool(credentialsService: CredentialsService)(implicit setting
...
+      config.setPassword(password)
+      val pool = new HikariDataSource(config)
+      pool
     }
   }
 
+  private val dbCache: LoadingCache[AuthenticatedUser, HikariDataSource] = CacheBuilder
+    .newBuilder()
+    .maximumSize(connectionCacheSize)
+    .expireAfterAccess(connectionCachePeriod)
```
bgaidioz, line 70: A question about potential leaks: if many different users use SQL, the `LoadingCache` reaches its configured maximum size. If an extra `HikariDataSource` is added to it, it leads to an existing `HikariDataSource` to be taken out. After that eviction, the `HikariDataSource` object isn't referenced anymore. It can be garbage collected. If it is garbage collected early enough, it won't close connections after their configured `idleTimeout` expires. The object doesn't exist anymore at that moment.

I read [in the docs](https://guava.dev/releases/19.0/api/docs/com/google/common/cache/CacheBuilder.html#removalListener), that's the `CacheBuilder` has a method `.removalListener` that permits to specify a callback to be executed upon deletion. That would certainly permit to perform the call to `.close` and avoid that.

bgaidioz: That would also be an opportunity to log a symmetric message about the connection pool being closed.

bgaidioz (approved):

PR #452, opened 2024-06-26
SQL client uses test containers + JDBC URL added to ProgramEnvironment

bgaidioz:
The changeset includes:
* SQL client tests now using a Postgres test container. That permits the CI to run these tests (they are disabled otherwise).
* Added `jdbcUrl` to the `ProgramEnvironment` so that the SQL compiler can just retrieve it before running a query or anything.
* removed dependencies of `sqlClient` on `raw.snapi.frontend` (removal of `CredentialsService` use) and `raw.sources` that was there by copy-paste accident.


sql-client/src/main/scala/raw/client/sql/SqlCompilerService.scala
```diff
@@ -18,23 +18,30 @@ import raw.client.api._
...
+  // A short lived database metadata (schema/table/column names) indexed by JDBC URL.
   private val metadataBrowsers = {
-    val loader = new CacheLoader[AuthenticatedUser, UserMetadataCache] {
-      override def load(user: AuthenticatedUser): UserMetadataCache =
-        new UserMetadataCache(user, connectionPool, settings)
+    val loader = new CacheLoader[String, UserMetadataCache] {
+      override def load(jdbcUrl: String): UserMetadataCache = new UserMetadataCache(
+        jdbcUrl,
+        connectionPool,
+        maxSize = settings.getInt("raw.client.sql.metadata-cache.max-matches"),
```
miguelbranco80, line 42: Just to ead settings once and store in private val, not everytime.

miguelbranco80 (approved):
Nice 👍

sql-client/src/main/scala/raw/client/sql/SqlCompilerService.scala
bgaidioz: Right.

PR #455, opened 2024-06-28
refactor(ci): get out of container

datYori:
- improve custom tasks using java and javac bins 
- run CI outside of container

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -195,7 +195,7 @@ lazy val snapiTruffle = (project in file("snapi-truffle"))
       println("Running Java annotation processor")
 
       val javaHomeDir = javaHome.value.getOrElse(sys.error("JAVA_HOME is not set"))
-      val javacExecutable = javaHomeDir / "bin" / "java"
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -188,6 +194,8 @@ lazy val snapiTruffle = (project in file("snapi-truffle"))
     runJavaAnnotationProcessor := {
       println("Running Java annotation processor")
 
+      val javaHomeDir = javaHome.value.getOrElse(sys.error("JAVA_HOME is not set"))
+      val javacExecutable = javaHomeDir / "bin" / "javac"
```
bgaidioz, line 198: Yes!

Some weeks ago, one of us couldn't compile because system wide `javac` wasn't the right version.

bgaidioz (approved):

build.sbt
datYori: with this patch, as long as JAVA_HOME is properly set (which the case for example when you do `sdk use java 21.0.1-graalce`), you are fine

PR #460, opened 2024-07-02
RD-11102: Added getUserJdbc to the Credentials API

bgaidioz:


sources/src/main/scala/raw/creds/api/CredentialsService.scala
```diff
@@ -241,11 +241,20 @@ trait CredentialsService extends RawService {
...
-   * automatically generated. This call is meant to be used by the SQL compiler in order to execute user-defined SQL queries.
+   * Returns a PostgreSQL FDW database for the given user, creating the database
+   * if it does not already exist.
    *
-   * @param user the Authenticated User which corresponds to the combination of RAW Organization and Repository
-   * @return the database name the corresponds to the given user.
+   * @param user the AuthenticatedUser which corresponds to the combination of RAW Organization and Repository
+   * @return the database name that corresponds to the given user.
    */
   def getUserDb(user: AuthenticatedUser): String
```
bgaidioz, line 250: getUserDb was left because it's used in a credentials test.

miguelbranco80 (approved):

PR #464, opened 2024-07-05
RD-11080: SqlCompiler crashes when a query is made of 100% comments

bgaidioz:
What is to be discussed here is: how to report that error? Technically the whole source code is failed, because no executable is found (that is the current patch). But underlining the whole code as people peacefully type a comment at the top, isn't convenient. Also, even an empty file should trigger an error?

Or can we underline the _end_ (last character) saying the SELECT statement is missing.

Or can we not report anything as a failure, e.g. `GetProgramDescription` would contain no errors, but also no "final executable", and fail only upon running? (Just thinking of this now.)


torcato (approved):
LGTM

Alexzerntev (approved):

PR #465, opened 2024-07-05
RD-11232 sql compiler crash parser computes a wrong line col

Alexzerntev:
- Fixed by striping trail and removing the extra new line

### DIFF
--- a/sql-client/src/main/scala/raw/client/sql/antlr4/RawSqlSyntaxAnalyzer.scala
+++ b/sql-client/src/main/scala/raw/client/sql/antlr4/RawSqlSyntaxAnalyzer.scala
@@ -26,9 +26,8 @@ class RawSqlSyntaxAnalyzer(val positions: Positions) extends Parsers(positions) 
     val source = StringSource(s)
     val rawErrorListener = new RawSqlErrorListener()
 
-    val striped = s.stripTrailing()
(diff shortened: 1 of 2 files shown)


sql-client/src/main/scala/raw/client/sql/antlr4/RawSqlSyntaxAnalyzer.scala
```diff
@@ -26,7 +26,8 @@ class RawSqlSyntaxAnalyzer(val positions: Positions) extends Parsers(positions)
     val source = StringSource(s)
     val rawErrorListener = new RawSqlErrorListener()
 
-    val lexer = new PsqlLexer(CharStreams.fromString(s))
+    val striped = s.stripTrailing().stripSuffix("\r\n").stripSuffix("\n").stripSuffix("\r")
```
bgaidioz, line 29: There's no risk there's another token later and we start to shift offsets?

Ah it's the whole code that gets stripped?

torcato, line 29: The stripTrainling should remove all white space, no?
Why does it need the stripSuffix's after?

torcato (approved):
I think stripSuffix's are unnecessary after the stripTrailing

PR #469, opened 2024-07-26
RD-11521: report connection failures to the user

bgaidioz:
Before the pool was global, the code logic was that a connection failure to FDW would be an internal bug, FDW should always be available, and the connection pool should be provisioned to match the expected usage.

To review carefully if this behavior is what we want, and is implemented properly: To report connection failures to the user, we should isolate the connection failure `SQLException` (those are identified with code `08xxx`), and process these ones as errors that are normal to hit sometimes:
* for `execute`, `validate` and `getProgramDescription`, they are caught are the message is reported (as a `String` or `ErrorMessage` with positions when needed). The `validate` and `getProgramDescription` need positions, I've put positions spanning the full code.
* for `hover`, we probably should ignore it and and do not report type info,
* for connection errors occuring in metadata caches, we do not report the error to the user (I think), and since we can't connect, store an empty list of matches. That entry will be wiped after the usual timeout.

The PR only adds tests to `execute`,`validate` and `getProgramDescription` and LSP calls.

### DIFF
--- a/sql-client/src/main/scala/raw/client/sql/SqlCompilerService.scala
+++ b/sql-client/src/main/scala/raw/client/sql/SqlCompilerService.scala
@@ -21,7 +21,7 @@ import raw.client.sql.writers.{TypedResultSetCsvWriter, TypedResultSetJsonWriter
 import raw.utils.{RawSettings, RawUtils}
 
 import java.io.{IOException, OutputStream}
-import java.sql.ResultSet
+import java.sql.{ResultSet, SQLException}
 import scala.util.control.NonFatal
 
 /**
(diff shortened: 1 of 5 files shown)


miguelbranco80:
One comment is that we found a slightly different behaviour when doing LSP calls vs validate/execute calls. So we should indeed be sure for the behaviour we want for LSP. Probably it is indeed safe to ignore as you mention for the hover.

sql-client/src/main/scala/raw/client/sql/metadata/UserMetadataCache.scala
```diff
@@ -132,4 +147,13 @@ class UserMetadataCache(jdbcUrl: String, connectionPool: SqlConnectionPool, maxS
     dotCompletionCache.get(seq).map(i => (i.name, i.tipe))
   }
 
+  private def isConnectionFailure(ex: SQLException) = {
+    val state = ex.getSQLState
+    if (state != null && state.startsWith("08")) {
+      logger.warn("SqlConnectionPool connection failure", ex)
```
miguelbranco80, line 153: I suggest you remove this logger.warn since you are logging in every case ex: SqlException if isConnecitonFailure.
It makes sense to me to log in the catch,  not here.

sql-client/src/main/scala/raw/client/sql/SqlConnectionPool.scala
```diff
@@ -146,7 +146,7 @@ class SqlConnectionPool()(implicit settings: RawSettings) extends RawService wit
         if (maybeConn.isEmpty && conns.size >= maxConnectionsPerDb) {
           // No connection was available to borrow, and too many being used for this db location, so we cannot open
           // any more!
-          throw new SQLException("too many connections active")
+          throw new SQLException("too many connections active", "08000")
```
miguelbranco80, line 149: Please explain why/what is 08000

sql-client/src/main/scala/raw/client/sql/SqlConnectionPool.scala
```diff
@@ -165,7 +165,7 @@ class SqlConnectionPool()(implicit settings: RawSettings) extends RawService wit
 
       // We could not successfully release any connection, so bail out.
       if (getTotalActiveConnections() >= maxConnections) {
-        throw new SQLException("no connections available")
+        throw new SQLException("no connections available", "08000")
```
miguelbranco80, line 168: Please explain why/what is 08000

sql-client/src/main/scala/raw/client/sql/SqlCompilerService.scala
```diff
@@ -421,4 +445,13 @@ class SqlCompilerService()(implicit protected val settings: RawSettings) extends
     rowAttrTypes.right.map(attrs => RawIterableType(RawRecordType(attrs.toVector, false, false), false, false))
   }
 
+  private def isConnectionFailure(ex: SQLException) = {
+    val state = ex.getSQLState
+    if (state != null && state.startsWith("08")) {
```
miguelbranco80, line 450: Mention here that 08 comes from SqlConnectionPool

sql-client/src/main/scala/raw/client/sql/metadata/UserMetadataCache.scala
```diff
@@ -132,4 +147,13 @@ class UserMetadataCache(jdbcUrl: String, connectionPool: SqlConnectionPool, maxS
     dotCompletionCache.get(seq).map(i => (i.name, i.tipe))
   }
 
+  private def isConnectionFailure(ex: SQLException) = {
+    val state = ex.getSQLState
+    if (state != null && state.startsWith("08")) {
```
miguelbranco80, line 152: Mention here that 08 comes from SqlConnectionPool

sql-client/src/test/scala/raw/client/sql/TestSqlConnectionFailures.scala
```diff
@@ -0,0 +1,582 @@
...
+import java.util.concurrent.{Executors, TimeUnit}
+import scala.io.Source
+
+class TestSqlConnectionFailures
+    extends RawTestSuite
+    with ForAllTestContainer
+    with SettingsTestContext
+    with TrainingWheelsContext {
+
+  /* The test suite triggers connection failures for both 'no connections
```
miguelbranco80, line 33: Comment with // instead

sql-client/src/test/scala/raw/client/sql/TestSqlConnectionFailures.scala
```diff
@@ -0,0 +1,582 @@
...
+     available' (the pool can't open a new connection) and 'too many
+     connections active' (the user runs too many queries in parallel)
+     for all implemented calls: execute, validate, getProgramDescription,
+     hover, dotCompletion, wordCompletion. For each we use sequential
+     or parallel queries to exhaust the pool in some way and assert the
+     failure is hit as expected.
+   */
+
+  // number of users to run with, this permits to test errors that are hit
+  // when a single user exhausts its share.
```
miguelbranco80, line 43: // Number of users to run with. This allows testing errors that occur
// when a single user exhausts their allocated share.

miguelbranco80 (approved):

PR #470, opened 2024-07-29
Simplify RawValue

miguelbranco80:
* Remove "snapi-only" types from the raw interface (e.g. RawLocationType).
* Instead, use directly Rql2Value for the staged compiler.


snapi-frontend/src/main/scala/raw/compiler/rql2/StagedCompiler.scala
```diff
@@ -111,9 +137,133 @@ trait StagedCompiler {
...
+        while (it.hasIteratorNextElement) {
+          val v1 = it.getIteratorNextElement
+          seq.append(polyglotValueToRql2Value(v1, innerType))
+        }
+        if (it.canInvokeMember("close")) {
+          val callable = it.getMember("close")
+          callable.execute()
+        }
+        Rql2IterableValue(seq)
+      case Rql2OrType(tipes, _) =>
```
miguelbranco80, line 215: Is this one correct, @bgaidioz ?

bgaidioz: It's reading private fields instead of calling getters. I could trigger the bug and will fix it in my PR.

bgaidioz (approved):

PR #475, opened 2024-08-05
RD-12433: Test that assert Http URL parsing is correctly handling arguments

bgaidioz:
The added `HttpPackageTest` tests passing args are passing. Because we pass the URL to a library which figures out. The most challenging one is where half of the args is passed in the URL, and the other half as a formal list of arguments, they're merged because the library sorts it out.

Hence I remove the "TODO" comment that was in the code since there's nothing to do (?)

Also added a CSV test making sure we correctly handle a mistakenly passed directory (in place of a file).

### DIFF
--- a/snapi-compiler/src/test/scala/com/rawlabs/snapi/compiler/tests/builtin/CsvPackageTest.scala
+++ b/snapi-compiler/src/test/scala/com/rawlabs/snapi/compiler/tests/builtin/CsvPackageTest.scala
@@ -34,6 +34,8 @@ class CsvPackageTest extends SnapiTestContext {
     |2|20|200
     |3|30|300""".stripMargin)
 
+  private val directory = headerLessData.getParent
+
   test(
     """Csv.Parse("1;tralala\n12;ploum\n3;ploum;\n4;NULL", type collection(record(a: int, b: string, c: string, d: string, e: string)),
       |skip = 0, delimiter = ";", nulls=["NULL", "12"])""".stripMargin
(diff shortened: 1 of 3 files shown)


snapi-client/src/test/scala/raw/compiler/rql2/tests/builtin/credentials/MySQLPackageTest.scala
```diff
@@ -169,6 +169,13 @@ import raw.testing.tags.TruffleTests
       |)""".stripMargin
   )(it => it should runErrorAs(s"""unknown database credential: ${mysqlCreds.database}""".stripMargin))
 
+  test(
```
bgaidioz, line 172: No. When wrapped in a list, the error shouldn't escape.

snapi-client/src/test/scala/raw/compiler/rql2/tests/regressions/RD10194Test.scala
```diff
@@ -29,7 +29,7 @@ import raw.testing.tags.TruffleTests
       |        where_clause = " where true " + actor_clause + title_clause + description_clause + category_clause + rating_clause,
       |        full_query = "select * from nicer_but_slower_film_list "+where_clause
       |    in
-      |        full_query
+      |        [List.From(hello)]
```
bgaidioz, line 32: Just to see what happens.

snapi-client/src/test/scala/raw/compiler/rql2/tests/regressions/RD10194Test.scala
```diff
@@ -27,9 +27,24 @@ import raw.testing.tags.TruffleTests
...
+      |                            category: string,
+      |                            price: decimal,
+      |                            length: short,
+      |                            rating: string,
+      |                            actors: string,
+      |                        )
+      |                    ))
       |    in
-      |        full_query
+      |        [List.From(hello)]
```
bgaidioz, line 47: Just to see the results.

snapi-client/src/test/scala/raw/compiler/rql2/tests/builtin/credentials/MySQLPackageTest.scala
bgaidioz, line 172: No. This test shouldn't fail, because the error object obtained from List.From, is wrapped in a list. We should get a list of one error.

PR #478, opened 2024-08-06
RD-12436 Adding examples for snowflake options, RD-12435 Add documentation to Location.String

torcato:
screenshots of the doc changes
image

image


snapi-frontend/src/main/scala/raw/compiler/rql2/builtin/SnowflakePackage.scala
```diff
@@ -102,11 +102,12 @@ class SnowflakeInferAndReadEntry extends SugarEntryExtension {
       ParamDoc(
         "options",
         typeDoc = TypeDoc(List("list")),
```
miguelbranco80, line 104: Is saying just "list" how we typically do it?

torcato: Yes, the other entries with similar arguments, like HttpGet or AwsV4SignedRequest have the same:
```scala
 ParamDoc(
        "args",
        TypeDoc(List("list")),
        """The query parameters arguments for the HTTP request, e.g. `[{"name", "john"}, {"age", "22"}]`. They are URL-encoded automatically.""".stripMargin,
        isOptional = true
      ),
      ParamDoc(
        "headers",
        TypeDoc(List("list")),
        """The HTTP headers to include in the request, e.g. `[{"Authorization", "Bearer 1234"}, {"Accept", "application/json"}]`.""".stripMargin,
        isOptional = true
      ),
```

miguelbranco80 (approved):

PR #499, opened 2024-08-30
RD-14664: SQL parser slow against large queries

bgaidioz:
The slow part was identified as parsing lists (the comma separated ones at least), which would lead to a nested cascade of lists of two elements: (1, (2, (3, (4,  ...)))) which eventually makes the parser very slow, instead of a single list of items 1,2,3,4,....

I attach two queries, the one used as a template we advertise in the web app, because it's also slow to parse, although it takes less than a second, and the large one that takes minutes. With the patch, both parse much faster (far less than a second).


Alexzerntev (approved):
The only case that this won't work is `,2,3,4`, this shouldn't parse, but will. But we have a safety net, the pssql call which will output an error. Will fix the parser in a next PR

PR #514, opened 2024-09-17
RD-14762: SqlCompilerService crashes on default null date parameter

bgaidioz:
Same for `TIME` and `TIMESTAMP`. Other types wouldn't crash, but they end up as e.g. a `RawString` that embeds a `null` string, instead of being a `RawNull`.


torcato (approved):
LGTM

PR #528, opened 2024-10-10
Alternative to fix RD-14971: remove inferrer threads and make inferre…

miguelbranco80:
…r a singleton

### DIFF
--- a/snapi-frontend/src/main/scala/com/rawlabs/snapi/frontend/base/CompilerContext.scala
+++ b/snapi-frontend/src/main/scala/com/rawlabs/snapi/frontend/base/CompilerContext.scala
@@ -12,24 +12,118 @@
 
 package com.rawlabs.snapi.frontend.base
 
+import com.google.common.base.Stopwatch
+import com.google.common.cache.{CacheBuilder, CacheLoader, LoadingCache}
 import com.rawlabs.utils.core.{RawSettings, RawUid}
 import com.typesafe.scalalogging.StrictLogging
-import com.rawlabs.snapi.frontend.inferrer.api.{InferrerInput, InferrerOutput, InferrerService}
+import com.rawlabs.snapi.frontend.inferrer.api.{InferrerException, InferrerInput, InferrerOutput, InferrerService}
+import com.rawlabs.snapi.frontend.snapi.extensions.Arg
 import com.rawlabs.utils.core._
+
+import java.util.concurrent.{ExecutionException, Executors, TimeUnit, TimeoutException}
+
+object CompilerContext {
+  private val INFERRER_TIMEOUT = "raw.snapi.frontend.inferrer.timeout"
+  private val INFERRER_EXPIRY = "raw.snapi.frontend.inferrer.expiry"
+
+  private val INFERRER_CACHE_SIZE = "raw.snapi.frontend.inferrer.cache-size"
+}
 
 /**
  * Contains state that is shared between different programs.
  */
 class CompilerContext(
-    val language: String,
     val user: RawUid,
     val inferrer: InferrerService
 )(
     implicit val settings: RawSettings
-) extends StrictLogging {
+) extends RawService
+    with StrictLogging {
 
-  def infer(properties: InferrerInput): Either[String, InferrerOutput] = {
-    inferrer.inferWithExpiry(properties)
+  import CompilerContext._
+
+  private val inferrerTimeoutMillis = settings.getDuration(INFERRER_TIMEOUT).toMillis
+  private val inferrerExpirySeconds = settings.getDuration(INFERRER_EXPIRY).toSeconds
+
+  private val inferrerCacheSize = settings.getInt(INFERRER_CACHE_SIZE)
+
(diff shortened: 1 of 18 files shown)


snapi-frontend/src/main/scala/com/rawlabs/snapi/frontend/snapi/ProgramContext.scala
```diff
@@ -33,11 +33,14 @@ class ProgramContext(
   private val stageCompilerCache = new mutable.HashMap[SnapiProgram, Either[ErrorCompilerMessage, SnapiValue]]
 
   def infer(
-      inferrerProperties: InferrerInput
+      inferrerProperties: InferrerInput,
+      mandatoryArgs: Seq[Arg],
```
miguelbranco80, line 37: This is not enough. It needs to distinguish csv from json calls and so forth. So it needs at least one extra 'string ' argument with that.

snapi-frontend/src/main/scala/com/rawlabs/snapi/frontend/base/CompilerContext.scala
```diff
@@ -12,24 +12,118 @@
...
+ * Key for the inferrer cache.
+ * The key is based on the arguments and properties passed to the inferrer.
+ * The properties are not considered in the equals and hashCode methods but only used to actually infer the output.
+ *
+ * @param mandatoryArgs the mandatory arguments
+ * @param optionalArgs the optional arguments
+ * @param varArgs the varargs
+ * @param inferrerInput the inferrer input
+ */
+private case class CompilerContextInferrerKey(
```
bgaidioz, line 111: It could be a real `class` to not confuse anyone with the usual way a `case class` is equal to another one.

snapi-frontend/src/main/scala/com/rawlabs/snapi/frontend/base/CompilerContext.scala
```diff
@@ -12,24 +12,118 @@
...
+    mandatoryArgs: Seq[Arg],
+    optionalArgs: Seq[(String, Arg)],
+    varArgs: Seq[Arg],
+    inferrerInput: InferrerInput
+) {
+  // Override equals and hashCode to consider only the arguments, not the properties
+  override def equals(other: Any): Boolean = other match {
+    case that: CompilerContextInferrerKey => this.mandatoryArgs == that.mandatoryArgs &&
```
bgaidioz, line 119: When the first are is a `location` (which it is in general, even when we pass a URL, it's turned into a `location`), the object we get is a `SnapiLocationValue`, which contains a `Location`. Which is the `class` that doesn't have a working comparison, the one used in `InferrerInput` that makes it not comparable.

snapi-truffle/src/main/java/com/rawlabs/snapi/truffle/SnapiLanguageCache.java
```diff
@@ -18,55 +18,27 @@
...
-    return map.computeIfAbsent(
-        user,
-        k -> {
-          InferrerService inferrer = InferrerServiceProvider.apply(rawSettings);
-          CompilerContext compilerContext =
-              new CompilerContext("snapi", user, inferrer, rawSettings);
-          return new Value(compilerContext, inferrer);
-        });
+  public InferrerService getInferrer(RawUid user, RawSettings rawSettings) {
+    return inferrerServiceCache.computeIfAbsent(user, k -> InferrerServiceProvider.apply(rawSettings));
```
bgaidioz, line 36: This is creating an inferrer with specific `rawSettings`, and store it for key `user`. Later even if passing other `rawSettings`, we'll find the one stored for the `user` key and return it (the one that has former settings).

snapi-frontend/src/main/scala/com/rawlabs/snapi/frontend/base/CompilerContext.scala
```diff
@@ -12,24 +12,82 @@
...
+  import CompilerContext._
+
+  private val inferrerTimeoutMillis = settings.getDuration(INFERRER_TIMEOUT).toMillis
+  private val inferrerExpirySeconds = settings.getDuration(INFERRER_EXPIRY).toSeconds
+
+  private val inferrerCacheSize = settings.getInt(INFERRER_CACHE_SIZE)
+
+  private val inferrerThreadPool = Executors.newCachedThreadPool(RawUtils.newThreadFactory("inferrer-thread"))
+
+  private val inferCache: LoadingCache[InferrerInput, Either[String, InferrerOutput]] = CacheBuilder
```
bgaidioz, line 51: The inference cache is in the `CompilerContext` now. And I remember that's the one which is created once and for all per user, right? (Maybe a second one is created for nested staged compilations, but that wouldn't another story). That ensures the cache never goes away.

The thread pool being a `CachedThreadPool` lets it grow ad infinitum but threads are eventually deleted (one minute by default). Under normal behavior we should see the number of inferrer threads being stable (proportional to the number of active users basically).

bgaidioz (approved):

PR #532, opened 2024-11-13
Fixed RD-15091: Ignore postgres offset when it is zero

bgaidioz:
When the offset is zero, it means there's no offset.


torcato (approved):
We looked at the code together. And added the test

PR #537, opened 2024-11-20
RD-15131: CRUD support in SqlCompilerService

bgaidioz:


torcato (approved):
We looked at it together

